PR #2, opened 2018-07-12
Initial setup -- adds test cases app and a couple of sample components to get started

danswick:
image


.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+	"*.js": ["eslint", "prettier --write", "git add"]
```
davidtheclark: Could we make these JSON files use the same indentation and have a line at the end? Sublime Text is letting you down.

package.json
```diff
@@ -0,0 +1,66 @@
...
+    "eslint": "^5.1.0",
+    "eslint-config-prettier": "^2.9.0",
+    "eslint-plugin-import": "^2.13.0",
+    "eslint-plugin-node": "^6.0.1",
+    "eslint-plugin-promise": "^3.8.0",
+    "eslint-plugin-react": "^7.10.0",
+    "globby": "^8.0.1",
+    "husky": "^0.14.3",
+    "lint-staged": "^7.2.0",
+    "lodash": "^4.17.10",
```
davidtheclark: Is lodash used anywhere?

package.json
```diff
@@ -0,0 +1,66 @@
...
+    "nanohref": "^3.0.3",
+    "nanorouter": "^3.1.1",
+    "pascal-case": "^2.0.1",
+    "pify": "^3.0.0",
+    "prettier": "^1.13.7",
+    "prop-types": "^15.6.2",
+    "react": "^16.4.1",
+    "react-dom": "^16.4.1",
+    "react-test-renderer": "^16.4.1"
```
davidtheclark: Looks like there a number of `devDependencies` that can be removed. The mapbox-react-components one these are from needs a little pruning, too, but it has tests and docs that this repo doesn't have, so not all its dependencies are relevant.

src/.eslintrc.json
```diff
@@ -0,0 +1,32 @@
...
+    "@mapbox/eslint-config-mapbox/promise",
+    "prettier"
+  ],
+  "rules": {
+    "prefer-const": "off",
+
+    "react/no-deprecated": "off",
+    "react/no-string-refs": "off",
+    "react/no-access-state-in-setstate": "off",
+    "react/jsx-boolean-value": "off"
```
davidtheclark: Let's remove these `rules` exceptions. I put them in mapbox-react-components because it wasn't quite ready for all the eslint-config-mapbox rules and I figured we could gradually fix things. For this repo, we could start with all the rules.

src/test-cases-app/dist/media-queries.json
```diff
@@ -0,0 +1,5 @@
+{
+  "--xl-screen": "screen and (min-width: 1200px)",
+  "--l-screen": "screen and (min-width: 800px)",
+  "--m-screen": "screen and (min-width: 640px)"
+}
```
davidtheclark: `src/test-cases-app/dist` should be Git-ignored.

src/test-cases-app/index.html
```diff
@@ -0,0 +1,23 @@
...
+  <script id="assembly-js" src="dist/assembly.js"></script>
+  <style type="text/css">
+  	#app {
+  		padding: 2em;
+  	}
+  	#app h1 {
+  		font-size: 2em;
+  		font-weight: bold;
+  		margin-bottom: 16px;
+  	}
```
davidtheclark: Are these styles needed anymore?

src/test-cases-app/test-cases-app.js
```diff
@@ -0,0 +1,15 @@
...
+import componentIndex from './component-index';
+
+class App extends React.Component {
+  render() {
+    return <ReactTestKitchen componentIndex={componentIndex} />;
+  }
+}
+
+const container = document.createElement('div');
+container.setAttribute('id', 'app');
```
davidtheclark: No need for an ID if you're not using it.

src/test-cases-app/test-cases-app.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import ReactTestKitchen from '@mapbox/react-test-kitchen';
+import componentIndex from './component-index';
```
davidtheclark: You'll want to `// eslint-disable-line` this import in case ESLint runs at a point when that file isn't generated.

danswick:
Thanks for the review @davidtheclark! Comments should be addressed ✅

PR #5, opened 2018-07-12
Trial run: small reorg and new component

davidtheclark:
- I added a `components/` subdirectory in `src/`, to differentiate components more clearly from `test-cases-app`.
- I remove `src/test-cases-app/component-index.js` from the Git tree because it's automatically generated so shouldn't be part of commits.
- I added a new component, OverviewHeader. Its test cases look like this:

192 168 131 12_9966_overviewheader 1

@colleenmcginnis and @danswick for review please.


davidtheclark:
Good suggestion, @colleenmcginnis. I added a commit that makes more of the props optional (while retaining the layout).

PR #10, opened 2018-07-13
Add page-layout component

### LINKED ISSUE #6: Add `PageLayout` component
The `PageLayout` component creates a page with two columns &mdash; a narrower one for a table of contents and a wider one for any kind of content. In the current docs, the kind of table of contents that is plugged into the layout is determined by page path &mdash; it might make more sense to specify in the front matter of each `.js` or `.md`  page file?

### Example in action

Here's the `PageLayout` with a `TocMultiPage` table of contents in the iOS docs:

image

### Dependencies

This component currently relies on an external component (https://github.com/yahoo/react-stickynode) to stick the container where the table of contents plugs in. It also includes the `BackToTopButton` that appears on narrow screens. 

### Props

- `sectionTitle` is just a string that will be at the top of the narrow column (table of contents).
- `tocContent` is a component that will render the table of contents that makes sense for each page and plug it into the narrow column.
- any `children` will go into the wider column (main content).

### Things to discuss

Does it make sense to determine which table of contents should be used in the parent component (in most cases `PageShell`) instead of here? Should we include the `BackToTopButton` as a part of this component or add it as a child of this component? 

### Starter code

```js
import React from 'react';
import BackToTopButton from './back-to-top-button';
import Sticky from 'react-stickynode';
import _ from 'lodash';

class PageLayout extends React.Component {
  static propTypes = {
    tocContent: PropTypes.node,
    sectionTitle: PropTypes.string
  };

  constructor(props) {
    super(props);
    this.state = {
      height: 0
    };
  }

  componentDidMount() {
    this.throttledHandleWindowResize();
    window.addEventListener('resize', this.throttledHandleWindowResize);
  }

  componentWillUnmount() {
    window.addEventListener('resize', this.throttledHandleWindowResize);
  }

  throttledHandleWindowResize = _.throttle(() => {
    const height = document.body.clientHeight;
    this.setState({
      bottomBoundaryValue: height - 450
    });
  }, 200);

  render() {
    
    return (
      <div className="grid grid--gut36 mr-neg36 mr0-mm">
        <div className="col col--4-mm col--12 bg-gray-faint px0 mt0-mm mt60">
          <Sticky
            enabled={true}
            bottomBoundary={this.state.bottomBoundaryValue}
            innerZ={1}
            top={50}
            activeClass="bg-gray-faint shadow-darken-10"
          >
            <div className="txt-ms pt24-mm pt0 viewport-almost-mm scroll-auto mt-neg60 mt0-mm">
              <div className="txt-l color-blue txt-fancy ml36 mb12 block-mm none">
                {this.props.sectionTitle}
              </div>
              {this.props.tocContent}
            </div>
          </Sticky>
        </div>
        <div
          id="docs-content"
          className="col col--8-mm col--12 mt24 mb60 pr0-mm pr36 prose"
        >
          {this.props.children}
          <BackToTopButton />
        </div>
      </div>
    );
  }
}

export { PageLayout };
```

colleenmcginnis:
Closes #6 

Still to do: 
- [ ] how to test with react-stickynode (🕵️‍♀️ #9 )
- [ ] more detailed test cases?
- [ ] make padding-top a prop (no `TopbarSticker` on help pages)

So far: 

image
image

### DIFF
--- a/src/components/page-layout/__tests__/page-layout-test-cases.js
+++ b/src/components/page-layout/__tests__/page-layout-test-cases.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import PageLayout from '../page-layout';
+
+const testCases = {};
+
+testCases.basic = {
+  description: 'Basic',
+  component: PageLayout,
+  props: {
+    sidebarContent: <div>Some content</div>,
+    sectionTitle: 'Some title',
+    children: <div>Doc content</div>
+  }
+};
+
+testCases.color = {
+  description: 'Specify background color',
+  component: PageLayout,
+  props: {
+    sidebarContent: <div>Some content</div>,
+    sectionTitle: 'Some title',
+    sidebarColor: 'bg-white',
+    children: <div>Doc content</div>
+  }
+};
+
+export { testCases };
--- a/src/components/page-layout/page-layout.js
+++ b/src/components/page-layout/page-layout.js
@@ -1,0 +1,81 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Sticky from 'react-stickynode';
+import _ from 'lodash';
+
+class PageLayout extends React.Component {
+  constructor(props) {
+    super(props);
+    this.throttledHandleWindowResize = this.throttledHandleWindowResize.bind(
+      this
+    );
+    this.state = {
+      height: 0
+    };
+  }
+
+  componentDidMount() {
+    this.throttledHandleWindowResize();
+    window.addEventListener('resize', this.throttledHandleWindowResize);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.throttledHandleWindowResize);
+  }
+
+  throttledHandleWindowResize() {
+    _.throttle(() => {
+      const height = document.body.clientHeight;
+      this.setState({
+        bottomBoundaryValue: height - 450
+      });
+    }, 200);
+  }
+
+  render() {
+    const { props, state } = this;
+    return (
+      <div className="grid">
+        <div
+          className={`col col--4-mm col--12 ${
+            props.sidebarColor
+          } px0 mt0-mm mt60`}
+        >
+          <Sticky
+            enabled={true}
+            bottomBoundary={state.bottomBoundaryValue}
+            innerZ={1}
+            top={50}
+            activeClass="bg-gray-faint shadow-darken-10"
+          >
+            <div className="txt-ms pt24-mm pt0 viewport-almost-mm scroll-auto ml36 mt-neg60 mt0-mm">
+              <div className="txt-l color-blue txt-fancy mb12 block-mm none">
+                {props.sectionTitle}
+              </div>
+              {props.sidebarContent}
+            </div>
+          </Sticky>
+        </div>
+        <div
+          id="docs-content"
+          className="col col--8-mm col--12 mt24 mb60 pr0-mm pr36 pl36 prose"
+        >
+          {props.children}
+        </div>
+      </div>
+    );
+  }
+}
+
+PageLayout.propTypes = {
+  sidebarContent: PropTypes.any,
+  sectionTitle: PropTypes.string,
+  sidebarColor: PropTypes.string,
+  children: PropTypes.any
+};
+
+PageLayout.defaultProps = {
+  sidebarColor: 'bg-gray-faint'
+};
+
+export default PageLayout;
--- a/package-lock.json
+++ b/package-lock.json
@@ -629,8 +629,7 @@
     "asap": {
       "version": "2.0.6",
       "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
-      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=",
-      "dev": true
+      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY="
     },
     "asn1.js": {
       "version": "4.10.1",
@@ -2304,6 +2303,11 @@
         }
       }
     },
+    "classnames": {
+      "version": "2.2.6",
+      "resolved": "https://registry.npmjs.org/classnames/-/classnames-2.2.6.tgz",
+      "integrity": "sha512-JR/iSQOSt+LQIWwrwEzJ9uk0xfN3mTVYMwt1Ir5mUcSN6pU+V4zQFFaJsclJbPuAUQH+yfWef6tm7l1quW3C8Q=="
+    },
     "cli-boxes": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/cli-boxes/-/cli-boxes-1.0.0.tgz",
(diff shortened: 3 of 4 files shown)


src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  render() {
+    const { props, state } = this;
+    return (
+      <div className="grid">
+        <div
+          className={`col col--4-mm col--12 ${
+            props.sidebarColor
+          } px0 mt0-mm mt60`}
```
davidtheclark: With reusable components, it's usually preferable to leave out layout styling that determines the component's relation to others around it. Instead, the component can be indifferent to where it's put — doesn't make assumptions about its context — and the parent can figure out how to arrange it in context.

Do you think that general rule applies here, or this is a worthy exception?

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+        >
+          <Sticky
+            enabled={true}
+            bottomBoundary={state.bottomBoundaryValue}
+            innerZ={1}
+            top={props.sidebarContentStickyTop}
+            activeClass="bg-gray-faint shadow-darken-10"
```
davidtheclark: I don't see the shadow class doing anything?

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+          >
+            <div className="txt-ms pt24-mm pt0 viewport-almost-mm scroll-auto ml36 mt-neg60 mt0-mm">
```
davidtheclark: I don't think `txt-ms` is a standard Assembly class. Are you adding it via custom CSS?

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+        >
+          {props.children}
+        </div>
+      </div>
+    );
+  }
+}
+
+PageLayout.propTypes = {
+  sidebarContent: PropTypes.any,
```
davidtheclark: This could be `PropTypes.node` — which means "Anything that can be rendered by React." That way an array of strings or an object or Errors will be rejected.

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+  sidebarTitle: PropTypes.string,
+  sidebarColor: PropTypes.string,
+  sidebarContentStickyTop: PropTypes.number,
+  children: PropTypes.any
```
davidtheclark: This can also be `PropTypes.node`. Is it required?

src/components/page-layout/page-layout.js
davidtheclark: Is it required?

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+            enabled={true}
+            bottomBoundary={state.bottomBoundaryValue}
+            innerZ={1}
+            top={props.sidebarContentStickyTop}
+            activeClass="bg-gray-faint shadow-darken-10"
+          >
+            <div className="txt-ms pt24-mm pt0 viewport-almost-mm scroll-auto ml36 mt-neg60 mt0-mm">
+              <div className="txt-l color-blue txt-fancy mb12 block-mm none">
+                {props.sidebarTitle}
+              </div>
```
davidtheclark: If a `sidebarTitle` is not provided you'll end up with a 12px margin beneath nothing — so, just some weird empty space. 

So maybe `sidebarTitle` should be required or else we only render that styled wrapper div if it's provided?

src/components/page-layout/page-layout.js
```diff
@@ -0,0 +1,77 @@
...
+              {props.sidebarContent}
+            </div>
+          </Sticky>
+        </div>
+        <div
+          id="docs-content"
+          className="col col--8-mm col--12 mt24 mb60 pr0-mm pr36 pl36 prose"
```
davidtheclark: The inclusion of `prose` here is pretty heavy-handed, making this component less generally useful (e.g. I'm not sure if GL JS will want that). Could we remove it, and apply `prose` in some other way where it's needed?

src/components/page-layout/__tests__/page-layout-test-cases.js
```diff
@@ -0,0 +1,38 @@
...
+  component: PageLayout,
+  props: {
+    sidebarContent: <div>Some content</div>,
+    sidebarTitle: 'Some title',
+    sidebarColor: 'bg-white',
+    children: <div>Doc content</div>
+  }
+};
+
+export { testCases };
```
davidtheclark: I'm not able to verify that this component is working as expected with a narrow viewport. Do you think there's any way we can change these test cases to demonstrate that?

src/components/page-layout/page-layout.js
colleenmcginnis: I'm not sure about `worthy`, but here's the situation:

If `mt0-mm mt60` is removed, we'll need to figure out how to add it back or any page with a sticky top nav will look like this on mobile:

introduction___mapbox

instead of this:

introduction___mapbox

Does that seem to be a worthy exception @davidtheclark?

src/components/page-layout/page-layout.js
colleenmcginnis: I don't think it _has_ to be required, though it might be a bit silly to have a sidebar with no content.

src/components/page-layout/page-layout.js
davidtheclark: Hmm, I don't think I understand. Also don't know about the `px0` class?

src/components/page-layout/__tests__/page-layout-test-cases.js
colleenmcginnis: Can you check out the changes I made in https://github.com/mapbox/frontend-documentation/pull/10/commits/e6530595cca2a30a562be2370968a97e4ad921ec and see if it's any clearer? I'm not sure what you're hoping to see.

src/components/page-layout/page-layout.js
colleenmcginnis: I changed `sidebarColor` to `sidebarTheme` so it can include color and positioning. 🤷‍♀️

src/components/page-layout/__tests__/page-layout-test-cases.js
davidtheclark: jul-13-2018 16-48-07

That's what I see on a narrow screen ^^. Does that look right? I'm not sure if it's right or not.

colleenmcginnis: Oh I see. I made a prop in https://github.com/mapbox/frontend-documentation/commit/e6530595cca2a30a562be2370968a97e4ad921ec that let's you decide if you want the sidebar content to stack or not (test case one = `false`, test case two = `true`): 

image

I _think_ if the content is stacked then we want it to stick (at least in the current design that's what's happening on [this page](https://www.mapbox.com/ios-sdk/maps/overview/) on narrow screens).

colleenmcginnis: @davidtheclark I think this is the last thing to address before merging... what do you think?

davidtheclark: I'm not sure about it but think you could just move forward and see how things go 🤷‍♂️

colleenmcginnis: Hmmm. Ok. Once all the open PRs are merged, I'll mess around with it and see if it's going to do what I want it to do in context and then revisit this general component to see if I can make it a little clearer.

PR #11, opened 2018-07-13
Add NavigationAccordion and NavigationDropdown

### LINKED ISSUE #7: Add NavigationAccordion and NavigationDropdown component
Edit: `TocMultiPage` <--> `NavigationAccordion `, `TocDropdown` <--> `NavigationDropdown `

Here comes another double component issue. 👯‍♂️  There are two types of tables of contents (so far) used in documentation. This one, `TocMultiPage`, lists guides on that are on multiple pages. It also relies on a  `TocDropdown` component that appears on narrow screens. 

### Example in action

Here's an example from the Maps SDK for iOS docs:

image

### Dependencies

This component relies on the Icon mapbox-react-component and a `TocDropdown` component that puts the first-level items into a dropdown on narrow screens.

### Props

`TocMultiPage`:

- `currentPath` comes from using `withLocation` in the `PageShell` component. It's used here to determine which item in the table of contents is active (and expanded).
- `sidebarContent` is an object with all the items that should be included in the table of contents. It's very particular. There's a `firstLevelToc` object with the title and path for each separate guide and a `secondLevelToc` for the headings within the current active guide.  

`TocDropdown`:

- `currentPath` is used here to determine which item in the table of contents is active (and expanded).
- `dropdownOptions` is just the `firstLevelToc` items.

### Test cases

- basic
- section with no secondLevelToc items

### Things to discuss

Nothing yet. 

### Starter code

`TocMultiPage`:

```js
import React from 'react';
import classnames from 'classnames';
import Icon from '@mapbox/react-icon';
import TocDropdown from './toc-dropdown';

class TocMultiPage extends React.PureComponent {
  static propTypes = {
    currentPath: PropTypes.string,
    sidebarContent: PropTypes.shape({
      firstLevelToc: PropTypes.arrayOf(
        PropTypes.shape({
          title: PropTypes.string.isRequired,
          path: PropTypes.string.isRequired
        })
      ).isRequired,
      seccondLevelToc: PropTypes.arrayOf(
        PropTypes.shape({
          level: PropTypes.number.isRequired,
          slug: PropTypes.string.isRequired,
          text: PropTypes.string.isRequired
        })
      )
    })
  };

  render() {
    const { props } = this;
    const secondLevelContent =
      props.sidebarContent.secondLevelToc &&
      props.sidebarContent.secondLevelToc.map(secondLevelTocItem => {
        return (
          <li key={secondLevelTocItem.slug} className="mt6">
            <a
              href={`#${secondLevelTocItem.slug}`}
              className="color-blue-on-hover text-decoration-none unprose"
            >
              {secondLevelTocItem.text}
            </a>
          </li>
        );
      });
    const firstLevelContent = props.sidebarContent.firstLevelToc.map(
      (page, index) => {
        const title = page.title;
        let icon = null;
        const isActive = this.props.currentPath === page.path;
        const breakLineClasses = classnames('py3 ', {
          'border-t border--gray-light': index !== 0
        });
        const textClasses = classnames('pl12 py12 txt-bold txt-m flex-child', {
          'color-black pb3': isActive
        });
        const activeSectionClasses = classnames(
          'pl24-mm px0 block-mm none pr12',
          {
            'bg-lighten75': isActive
          }
        );
        if (!isActive) {
          icon = (
            <Icon
              name="chevron-down"
              className="flex-child flex-child--no-shrink icon color-gray h24 w24"
            />
          );
        }
        return (
          <div key={index} className={activeSectionClasses}>
            <div className={breakLineClasses}>
              <a
                href={page.path}
                className="color-blue-on-hover color-gray text-decoration-none unprose flex-parent flex-parent--space-between-main flex-parent--center-cross"
              >
                <div className={textClasses}>{title}</div>
                {icon}
              </a>
              {isActive && secondLevelContent ? (
                <div className="ml24 pt0">
                  <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover-anchor unprose">
                    {secondLevelContent}
                  </ul>
                </div>
              ) : (
                ''
              )}
            </div>
          </div>
        );
      }
    );
    return (
      <div>
        <div className="block-mm none">{firstLevelContent}</div>
        <div className="none-mm block bg-gray-faint shadow-darken10">
          <TocDropdown
            dropdownOptions={this.props.sidebarContent.firstLevelToc}
            currentPath={this.props.currentPath}
          />
        </div>
      </div>
    );
  }
}

export { TocMultiPage };
```

`TocDropdown`:

```js
import React from 'react';
import ControlSelect from '@mapbox/react-control-select';
import { routeToPrefixed } from '@mapbox/batfish/modules/route-to';

export default class TocDropdown extends React.Component {
  static propTypes = {
    dropdownOptions: PropTypes.arrayOf(
      PropTypes.shape({
        title: PropTypes.string.isRequired,
        path: PropTypes.string.isRequired
      })
    ).isRequired
  };

  render() {
    const optionsArray = this.props.dropdownOptions.map(option => {
      return {
        label: option.title,
        value: option.path
      };
    });
    let currentOption = optionsArray.filter(option => {
      return this.props.currentPath == option.value;
    });

    return (
      <div className="py24 px24 flex-parent flex-parent--space-between-main flex-parent--center-cross">
        <ControlSelect
          id="navigate-this-section"
          onChange={value => {
            routeToPrefixed(value);
          }}
          value={currentOption[0].value}
          options={optionsArray}
          themeControlSelect="select--stroke round-full bg-white"
          themeControlSelectContainer="flex-child flex-child--no-shrink"
        />
      </div>
    );
  }
}

```

colleenmcginnis:
Closes #7 

`NavigationAccordion`:

image

`NavigationDropdown`:

image

### DIFF
--- a/src/components/navigation-accordion/navigation-accordion.js
+++ b/src/components/navigation-accordion/navigation-accordion.js
@@ -1,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import Icon from '@mapbox/react-icon';
+import NavigationDropdown from '../navigation-dropdown/navigation-dropdown';
+
+class NavigationAccordion extends React.PureComponent {
+  render() {
+    const { props } = this;
+    const secondLevelContent =
+      props.contents.seccondLevelItems &&
+      props.contents.seccondLevelItems.map(item => {
+        return (
+          <li key={item.slug} className="mt6">
+            <a
+              href={`#${item.slug}`}
+              className="color-blue-on-hover text-decoration-none unprose"
+            >
+              {item.text}
+            </a>
+          </li>
+        );
+      });
+    const firstLevelContent = props.contents.firstLevelItems.map(
+      (page, index) => {
+        const title = page.title;
+        let icon = null;
+        const isActive = this.props.currentPath === page.path;
+        const breakLineClasses = classnames('py3 ', {
+          'border-t border--gray-light': index !== 0
+        });
+        const textClasses = classnames('pl12 py12 txt-bold txt-m flex-child', {
+          'color-black pb3': isActive
+        });
+        const activeSectionClasses = classnames(
+          'pl24-mm px0 block-mm none pr12',
+          {
+            'bg-lighten75': isActive
+          }
+        );
+        if (!isActive) {
+          icon = (
+            <Icon
+              name="chevron-down"
+              className="flex-child flex-child--no-shrink icon color-gray h24 w24"
+            />
+          );
+        }
+        return (
+          <div key={index} className={activeSectionClasses}>
+            <div className={breakLineClasses}>
+              <a
+                href={page.path}
+                className="color-blue-on-hover color-gray text-decoration-none unprose flex-parent flex-parent--space-between-main flex-parent--center-cross"
+              >
+                <div className={textClasses}>{title}</div>
+                {icon}
+              </a>
+              {isActive && secondLevelContent ? (
+                <div className="ml24 pt0">
+                  <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover-anchor unprose">
+                    {secondLevelContent}
+                  </ul>
+                </div>
+              ) : (
+                ''
+              )}
+            </div>
+          </div>
+        );
+      }
+    );
+    return (
+      <div>
+        <div className="block-mm none">{firstLevelContent}</div>
+        <div className="none-mm block bg-gray-faint">
+          <NavigationDropdown
+            currentPath={props.currentPath}
+            dropdownOptions={props.contents.firstLevelItems}
+          />
+        </div>
+      </div>
+    );
+  }
+}
+
+NavigationAccordion.propTypes = {
+  currentPath: PropTypes.string,
+  contents: PropTypes.shape({
+    firstLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        title: PropTypes.string.isRequired,
+        path: PropTypes.string.isRequired
+      })
+    ).isRequired,
+    seccondLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        level: PropTypes.number.isRequired,
+        slug: PropTypes.string.isRequired,
+        text: PropTypes.string.isRequired
+      })
+    )
+  })
+};
+
+export default NavigationAccordion;
--- a/src/components/navigation-accordion/navigation-accordion.js
+++ b/src/components/navigation-accordion/navigation-accordion.js
@@ -8,15 +8,15 @@ class NavigationAccordion extends React.PureComponent {
   render() {
     const { props } = this;
     const secondLevelContent =
-      props.contents.seccondLevelItems &&
-      props.contents.seccondLevelItems.map(item => {
+      props.contents.secondLevelItems &&
+      props.contents.secondLevelItems.map(item => {
         return (
-          <li key={item.slug} className="mt6">
+          <li key={item.path} className="mb6">
             <a
-              href={`#${item.slug}`}
+              href={`#${item.path}`}
               className="color-blue-on-hover text-decoration-none unprose"
             >
-              {item.text}
+              {item.title}
             </a>
           </li>
         );
@@ -30,7 +30,7 @@ class NavigationAccordion extends React.PureComponent {
           'border-t border--gray-light': index !== 0
         });
         const textClasses = classnames('pl12 py12 txt-bold txt-m flex-child', {
-          'color-black pb3': isActive
+          'color-black': isActive
         });
         const activeSectionClasses = classnames(
           'pl24-mm px0 block-mm none pr12',
@@ -40,12 +40,22 @@ class NavigationAccordion extends React.PureComponent {
         );
         if (!isActive) {
           icon = (
-            <Icon
-              name="chevron-down"
-              className="flex-child flex-child--no-shrink icon color-gray h24 w24"
-            />
+            <div className="flex-child flex-child--no-shrink">
+              <Icon name="chevron-down" className="icon color-gray h24 w24" />
+            </div>
           );
         }
+        let renderedSecondLevelContent = '';
+        if (isActive && secondLevelContent) {
+          renderedSecondLevelContent = (
+            <div className="ml24 pt0">
+              <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover unprose">
+                {secondLevelContent}
+              </ul>
+            </div>
+          );
+        }
+
         return (
           <div key={index} className={activeSectionClasses}>
             <div className={breakLineClasses}>
@@ -56,15 +66,7 @@ class NavigationAccordion extends React.PureComponent {
                 <div className={textClasses}>{title}</div>
                 {icon}
               </a>
-              {isActive && secondLevelContent ? (
-                <div className="ml24 pt0">
-                  <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover-anchor unprose">
-                    {secondLevelContent}
-                  </ul>
-                </div>
-              ) : (
-                ''
-              )}
+              {renderedSecondLevelContent}
             </div>
           </div>
         );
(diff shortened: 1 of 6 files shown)


src/components/navigation-dropdown/navigation-dropdown.js
```diff
@@ -0,0 +1,46 @@
...
+    const currentOption = optionsArray.filter(option => {
+      return props.currentPath === option.value;
+    });
+
+    return (
+      <div className="py24 px24 flex-parent flex-parent--space-between-main flex-parent--center-cross">
+        <ControlSelect
+          id="navigate-this-section"
+          onChange={() => {
+            // routeToPrefixed(value);
```
colleenmcginnis: @davidtheclark suggestions on how to handle this here? I'm guessing we don't want to use something Batfish-specific like this, but let me know if it's ok!

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
...
+          'pl24-mm px0 block-mm none pr12',
+          {
+            'bg-lighten75': isActive
+          }
+        );
+        if (!isActive) {
+          icon = (
+            <Icon
+              name="chevron-down"
+              className="flex-child flex-child--no-shrink icon color-gray h24 w24"
```
davidtheclark: I believe it's more reliable for IE11 if you put the flexbox layout wrappers on a div outside the `<Icon>` instead of on the icon itself.

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
...
+              </a>
+              {isActive && secondLevelContent ? (
+                <div className="ml24 pt0">
+                  <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover-anchor unprose">
+                    {secondLevelContent}
+                  </ul>
+                </div>
+              ) : (
+                ''
+              )}
```
davidtheclark: I try to avoid putting ternaries within the JSX, because it can get illegible pretty quickly. I'd suggest either making a `renderSomething()` method on the class or assigning the value of this block to a variable somewhere above the return statement.

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
...
+              <a
+                href={page.path}
+                className="color-blue-on-hover color-gray text-decoration-none unprose flex-parent flex-parent--space-between-main flex-parent--center-cross"
+              >
+                <div className={textClasses}>{title}</div>
+                {icon}
+              </a>
+              {isActive && secondLevelContent ? (
+                <div className="ml24 pt0">
+                  <ul className="txt-m pb12 inline-block-mm none color-blue-on-hover-anchor unprose">
```
davidtheclark: I don't think `color-blue-on-hover-anchor` is part of Assembly?

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
...
+    firstLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        title: PropTypes.string.isRequired,
+        path: PropTypes.string.isRequired
+      })
+    ).isRequired,
+    seccondLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        level: PropTypes.number.isRequired,
+        slug: PropTypes.string.isRequired,
```
davidtheclark: Could we use `path` in the same way as for `firstLevelItems`? Or how about `url` or `href`? Not sure we *need* to build in the assumption that second-level items are always hash links.

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
...
+  contents: PropTypes.shape({
+    firstLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        title: PropTypes.string.isRequired,
+        path: PropTypes.string.isRequired
+      })
+    ).isRequired,
+    seccondLevelItems: PropTypes.arrayOf(
+      PropTypes.shape({
+        level: PropTypes.number.isRequired,
```
davidtheclark: I don't see this prop being used? Maybe I'm missing it?

src/components/navigation-dropdown/navigation-dropdown.js
```diff
@@ -0,0 +1,46 @@
...
+    const { props } = this;
+    const optionsArray = props.dropdownOptions.map(option => {
+      return {
+        label: option.title,
+        value: option.path
+      };
+    });
+    const currentOption = optionsArray.filter(option => {
+      return props.currentPath === option.value;
+    });
```
davidtheclark: `currentOption` is actually the first item in the array, so I might grab that here so the variable name is more clear

src/components/navigation-dropdown/navigation-dropdown.js
```diff
@@ -0,0 +1,46 @@
...
+
+    return (
+      <div className="py24 px24 flex-parent flex-parent--space-between-main flex-parent--center-cross">
```
davidtheclark: Are these flex classes necessary? Looks like there's only one flex-child?

src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import Icon from '@mapbox/react-icon';
+import NavigationDropdown from '../navigation-dropdown/navigation-dropdown';
+
+class NavigationAccordion extends React.PureComponent {
+  render() {
+    const { props } = this;
+    const secondLevelContent =
```
davidtheclark: Spacing looks a little off when there's no second-level content:

2018-07-13 at 2 47 pm

src/components/navigation-dropdown/navigation-dropdown.js
davidtheclark: NavigationDropdown can have `onChange` as a prop, and pass that the value of the option that was selected.

NavigationAccordion could take an `onDropdownChange` prop that passes through. 

It's kind of awkward, but would work in this context. It's arguable that `<select>`s have some undesirable qualities for usage as a navigation dropdown — so we might want to consider building a new component for this in the future.

src/components/navigation-accordion/navigation-accordion.js
colleenmcginnis: Sure. That's just the shape that's collected via Batfish. I suppose I could rename the names in the PageShell before passing here?

davidtheclark: Yeah, and when passing it in you'd also prefix the `#`.

src/components/navigation-dropdown/navigation-dropdown.js
colleenmcginnis: Must be leftover from an `earlier attempt`.

src/components/navigation-dropdown/navigation-dropdown.js
colleenmcginnis: I'm having a hard time understanding how information is flowing in that scenario ☝️

src/components/navigation-dropdown/navigation-dropdown.js
davidtheclark:
```js
const currentOption = optionsArray.filter(option => {
  return props.currentPath === option.value;
})[0];
```

That `[0]` at the end means `currentOption` will be the current option instead of an array with one item that is the current option.

davidtheclark: You *could* use `find` instead of `filter`, except find isn't supported on IE11 🤷‍♂️

colleenmcginnis: Oh I understand. Thanks!

src/components/navigation-dropdown/navigation-dropdown.js
```diff
@@ -0,0 +1,44 @@
...
+        label: option.title,
+        value: option.path
+      };
+    });
+    const currentOption = optionsArray.filter(option => {
+      return props.currentPath === option.value;
+    })[0];
+
+    return (
+      <div className="py24 px24">
```
davidtheclark: Not sure about this padding, since I think that layout should be determined by the parent element.

PR #14, opened 2018-07-13
Add ProductMenu, ProductMenuDropdown components. Add ProductMenuItems data.

danswick:
closes https://github.com/mapbox/frontend-documentation/issues/3

image

image

After chatting with @colleenmcginnis yesterday about the data that populates the ProductMenu, it seems clear that `frontend-documentation` would make a good home for the `truth` about which products should be in the menu. I included the most recent product list I could find in a new `data` directory. I'm not sure if that's the most sensible way to do it! 

I'm also unsure whether it would be better to refactor `ProductMenuDropdown` a bit so it accepts a prop with all of the product data.


src/components/product-menu-dropdown/__tests__/product-menu-dropdown-test-cases.js
```diff
@@ -0,0 +1,15 @@
+import { ProductMenuDropdown } from '../product-menu-dropdown';
```
davidtheclark: We should decide whether to favor default or named exports by default in the repo and stick with that. I'm fine with either way.

src/components/product-menu-dropdown/product-menu-dropdown.js
```diff
@@ -0,0 +1,88 @@
...
+    const activeDotStyle = {
+      content: '',
+      width: '6px',
+      height: '6px',
+      borderRadius: '3px',
+      background: '#4264fb',
+      position: 'absolute',
+      margin: '10px 0 0 -12px',
+      display: 'none'
+    };
```
davidtheclark: Since this object doesn't get affected by the props, it could be declared outside the component's `render` method, which means we don't have to re-create the object on every render.

src/components/product-menu-dropdown/product-menu-dropdown.js
```diff
@@ -0,0 +1,88 @@
...
+    const allCategories = this.props.categories.map((category, index) => {
+      const allProducts = category.products.map((product, index) => {
+        const liClasses = classnames('mt6 relative ', {
+          'txt-bold': product.url && location.pathname.indexOf(product.url) > -1
```
davidtheclark: `window.location` won't be available when pages are rendered in Node, so there's a chance the current logic will break things during static HTML rendering.

src/components/product-menu-dropdown/product-menu-dropdown.js
```diff
@@ -0,0 +1,88 @@
...
+        });
+        const dotClasses = classnames({
+          'inline-block': location.pathname.indexOf(product.url) > -1,
```
davidtheclark: You could consolidate this logic into a single variable: `isActive`. Then you only have to figure it out once and can reapply it in these various contexts.

src/components/product-menu-dropdown/product-menu-dropdown.js
```diff
@@ -0,0 +1,88 @@
...
+      const categoryClasses = classnames('wmin240 py24', {
+        'border-b border--gray-light': index < categoryLength - 1
+      });
+      return (
+        <div className={categoryClasses} key={index}>
+          <div className="txt-bold color-gray">
+            <Icon
+              name={category.icon}
+              inline={true}
+              className="icon mb-neg6 mr6 h20 w20 color-gray inline"
```
davidtheclark: Is `inline` necessary? Or is it covered by `inline={true}`?

src/components/product-menu/__tests__/product-menu-test-cases.js
```diff
@@ -0,0 +1,15 @@
+import { ProductMenu } from '../product-menu';
+
+const testCases = {};
+const noRenderCases = {};
```
davidtheclark: We could probably trim out `noRenderCases` when there aren't any, so there's less boilerplate.

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import PopoverTrigger from '@mapbox/react-popover-trigger';
+import Icon from '@mapbox/react-icon';
+import { ProductMenuDropdown } from '../product-menu-dropdown/product-menu-dropdown';
+// You could import this list from, say, a data selector if you like
+// On Mapbox docs pages, categories data is imported from a central location
+// to keep menus in sync across the site.
```
davidtheclark: That comment is no longer true, true, right?

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+import { ProductMenuItems } from '../../data/product-menu-items';
+
+const popoverProps = {
+  placement: 'bottom',
+  themePopover:
+    'round shadow-darken25 viewport-almost-but-not-always scroll-auto scroll-styled'
```
davidtheclark: `viewport-almost-but-not-always` is not in Assembly, right?

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+
+  render() {
+    return (
+      <PopoverTrigger
+        content={this.renderMenu}
+        popoverProps={popoverProps}
+        onPopoverOpen={this.onPopoverOpen}
+        onPopoverClose={this.onPopoverClose}
+      >
+        <div className="fl wmax240-ml wmax180-mm flex-parent flex-parent--space-between-main flex-parent--center-cross">
```
davidtheclark: Why is this floating left?

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+          <div className="flex-child inline-block txt-fancy txt-l cursor-pointer border-b border-b--2 border--white border--blue-on-hover txt-truncate">
+            {this.props.product} SDK for {this.props.platform}
+          </div>
+          <Icon
+            name="caret-down"
+            inline={true}
+            className="flex-child fr icon h30 w30"
```
davidtheclark: I think IE11 has problem with flex display on SVGs, so a wrapping div should take care of the layout instead of the Icon itself.

davidtheclark: Why is this floating right?

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+  render() {
+    return (
+      <PopoverTrigger
+        content={this.renderMenu}
+        popoverProps={popoverProps}
+        onPopoverOpen={this.onPopoverOpen}
+        onPopoverClose={this.onPopoverClose}
+      >
+        <div className="fl wmax240-ml wmax180-mm flex-parent flex-parent--space-between-main flex-parent--center-cross">
+          <div className="flex-child inline-block txt-fancy txt-l cursor-pointer border-b border-b--2 border--white border--blue-on-hover txt-truncate">
```
davidtheclark: flex-child and inline-block cancel each other out.

davidtheclark: Curious why this div has a pointer cursor, instead of the higher-level container? I notice on the live site that if you hover over the caret icon you don't see the cursor, but you *can* click — probably we could improve on that.

davidtheclark: `border-transparent` could be the default (non-hover), in case we ever put this on a non-white background.

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { clicked: false };
+
+    this.onPopoverOpen = this.onPopoverOpen.bind(this);
+    this.onPopoverClose = this.onPopoverClose.bind(this);
+  }
+
+  renderMenu() {
+    return <ProductMenuDropdown categories={ProductMenuItems} />;
```
danswick: Could we generalize this component a bit more by rendering `ProductMenu`'s children instead of `ProductMenuDropdown` specifically?

src/components/product-menu/product-menu.js
```diff
@@ -0,0 +1,66 @@
...
+          />
+        </div>
+      </PopoverTrigger>
+    );
+  }
+}
+
+ProductMenu.propTypes = {
+  platform: PropTypes.string.isRequired,
+  product: PropTypes.string.isRequired
```
davidtheclark: @colleenmcginnis Is there any chance that this menu might contain *non* "SDK"s at some point? If so, maybe we want to remove the logic that forces an `x SDK for y` format, and instead allow products to be freeform strings?

src/components/product-menu/product-menu.js
davidtheclark: I'm not sure it needs to be generalized, really, unless we decide at some point that we want to repeat the button style with different content. For now, these components are coupled immediately with the product names themselves, and I think that's ok.

danswick: @davidtheclark 🙈 you're so fast! whelp, I rearranged the furniture a bit and I think I like it better this way. I don't know that we'll have an immediate use case for this style of button with other stuff inside of it, but I can see it being useful in the future. 

Would you and @colleenmcginnis mind taking a peek at the new arrangement and let me know if you think it's worse? 

image

davidtheclark: 👍

src/components/product-menu-dropdown/__tests__/product-menu-dropdown-test-cases.js
danswick: Per chat, we're going to stick with `always named` for now.

src/components/product-menu/__tests__/product-menu-test-cases.js
danswick: 👍 i don't think we have any of these set up now, so I added a reminder for myself to add something to the README - https://github.com/mapbox/frontend-documentation/issues/22#issuecomment-405339252.

src/components/product-menu-dropdown/product-menu-dropdown.js
danswick: @davidtheclark looks like they're both needed: 

- `.inline` sets `display: inline !important`
- `inline={true}` does some math to position the icon vertically: https://github.com/mapbox/mapbox-react-components/blob/master/packages/react-icon/src/icon.js#L19-L28

src/components/product-menu/product-menu.js
danswick: per chat with @colleenmcginnis, we're going to move all of the docs-specific UI out of copypasta limbo and into this repo for docs-specific-components to reference.

src/components/product-menu/product-menu.js
danswick: I'm not sure and will defer to @colleenmcginnis here. Maybe it's there as a fallback for browsers that don't support flexbox?

src/components/product-menu/product-menu.js
danswick: > Why is this floating right?

Similar to above, I'm not totally sure. Would you mind hopping in here, @colleenmcginnis?

src/components/product-menu/product-menu.js
colleenmcginnis: I think you can 🔪 @danswick

src/components/product-menu/product-menu.js
colleenmcginnis: 🔪

src/components/product-menu/product-menu.js
colleenmcginnis: Yes. We should change that &mdash; it will break for GL JS. @danswick let's just use a prop like `productName` to pass a string with the exact name we want to display.

danswick:
Thanks for 👀 @colleenmcginnis. Updated per your comments. Regarding the menu overrunning the bottom of the page, yes, I believe this will be resolved once our custom docs classes are added.

PR #17, opened 2018-07-14
Card container

### LINKED ISSUE #13: Add `CardContainer` component
The `CardContainer` component organizes cards (examples, tutorials, etc.) into topics/sections and adds a title to the grouping of cards. 

### Example in action

image

### Dependencies

Should it include the `ExampleCard` (or `TutorialCard` or `DocCard`) or should an array of cards be passed as a prop? 

### Props

- `category` object with a `title` and `path`
- `cards` an array of all the cards for this category

### Things to discuss

n/a

### Starter code

Here's the `Example`

```js
import React from 'react';
import { ExampleCard } from './example-card';

class ExampleSection extends React.PureComponent {
  static propTypes = {
    category: PropTypes.shape({
      title: PropTypes.string.isRequired,
      path: PropTypes.string.isRequired
    }).isRequired,
    examples: PropTypes.array.isRequired
  };

  render() {
    const categoryID = this.props.category.path.split('#')[1];
    const allExampleCards = this.props.examples.map((example, index) => {
      return (
        <ExampleCard
          key={index}
          exampleTitle={example.title}
          exampleUrl={example.path}
          exampleDescription={example.description}
          exampleThumbnail={example.thumbnail}
        />
      );
    });
    return (
      <div>
        <a
          href={`${this.props.category.path}`}
          className="unprose mb12 block color-blue-on-hover"
        >
          <h2 className="pt60 txt-l txt-capitalize-first" id={categoryID}>
            {this.props.category.title} ({this.props.examples.length})
          </h2>
        </a>
        <div className="grid grid--gut12 card-container">{allExampleCards}</div>
      </div>
    );
  }
}

export { ExampleSection };
```

colleenmcginnis:
Closes #13 

This creates a `CardContainer` component and two kinds of cards: `CardWithImage` and `CardTextOnly` (these replace `TutorialCard`). A few questions: 

- Is it 🆒 to keep these as two separate card components or could they be combined in some way? 
- I'm trying to set up the thumbnail image to work with how we're using `AppropriateImage` in the Android, iOS, and Unity docs: 

```js
<AppropriateImage
  backgroundSize="100% auto"
  imageId={this.props.exampleThumbnail}
  background={true}
/>
```

Here's how it is looking so far: 

image


src/components/card-text-only/card-text-only.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class CardTextOnly extends React.Component {
+  render() {
+    const { props } = this;
+    return (
+      <div className="col col--12 mb12">
```
davidtheclark: I think that the parent component should determine this element's position in the context of a layout. Looks like that responsibility falls under CardContainer.

It may be that we could even eliminate this wrapper div.

src/components/card-container/card-container.js
```diff
@@ -0,0 +1,32 @@
...
+      <div>
+        <a
+          href={props.category.path}
+          className="unprose mb12 block color-blue-on-hover"
+        >
+          <h2 className="pt60 txt-l" id={categoryID}>
+            {props.category.title} ({props.cards.length})
+          </h2>
+        </a>
+        <div className="grid grid--gut12 card-container">{props.cards}</div>
```
davidtheclark: The `card-container` doesn't do anything, right?

src/components/card-container/card-container.js
```diff
@@ -0,0 +1,32 @@
...
+  render() {
+    const { props } = this;
+    const categoryID = props.category.path.split('#')[1];
+    return (
+      <div>
+        <a
+          href={props.category.path}
+          className="unprose mb12 block color-blue-on-hover"
+        >
+          <h2 className="pt60 txt-l" id={categoryID}>
```
davidtheclark: This component shouldn't determine its own top padding: the parent should lay it out in relation to other things.

src/components/card-container/card-container.js
```diff
@@ -0,0 +1,32 @@
...
+      </div>
+    );
+  }
+}
+
+CardContainer.propTypes = {
+  category: PropTypes.shape({
+    title: PropTypes.string.isRequired,
+    path: PropTypes.string.isRequired
+  }).isRequired,
```
davidtheclark: Not clear to me why `category` is an object, instead of `title` and `path` being props? Seems like it would make a simpler component API to flatten that.

src/components/card-container/card-container.js
```diff
@@ -0,0 +1,32 @@
...
+  cards: PropTypes.array.isRequired
```
davidtheclark: It's an `arrayOf(PropTypes.node)`, right?

src/components/card-text-only/card-text-only.js
```diff
@@ -0,0 +1,30 @@
...
+        <a
+          className="color-gray-dark transition shadow-darken10-on-hover round clip inline-block w-full px12 py12 unprose"
```
davidtheclark: Could this be `block` instead of `inline-block w-full`?

src/components/card-text-only/card-text-only.js
```diff
@@ -0,0 +1,30 @@
...
+          href={props.path}
+          style={{ verticalAlign: 'top' }}
+        >
+          <div className="px6 pb6">
```
davidtheclark: Why is there padding both on the div and the a? Could we remove the div?

src/components/card-with-image/card-with-image.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class CardWithImage extends React.Component {
```
davidtheclark: I agree with the idea in your comment that these two components could be effectively merged into one.

src/components/card-text-only/card-text-only.js
```diff
@@ -0,0 +1,30 @@
...
+
+class CardTextOnly extends React.Component {
+  render() {
+    const { props } = this;
+    return (
+      <div className="col col--12 mb12">
+        <a
+          className="color-gray-dark transition shadow-darken10-on-hover round clip inline-block w-full px12 py12 unprose"
+          href={props.path}
+          style={{ verticalAlign: 'top' }}
```
davidtheclark: What is this `verticalAlign` for?

src/components/card-with-image/card-with-image.js
```diff
@@ -0,0 +1,32 @@
...
+      </div>
+    );
+  }
+}
+
+CardWithImage.propTypes = {
+  title: PropTypes.string.isRequired,
+  path: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  thumbnail: PropTypes.node.isRequired
```
davidtheclark: Good move allowing the thumbnail to be passed in from the outside.

src/components/card-container/card-container.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+class CardContainer extends React.PureComponent {
```
davidtheclark: Are cards *with* images ever going to coexist with cards *without* images? Would that work, or would it break the layout?

colleenmcginnis:
@davidtheclark I addressed your comments and combined the two card types into one. I also added an `ExamplesPage` component that can be used to make consistent how the `CardContainer`s should be arranged on SDK Examples pages (this makes me feel better about addressing comments like this https://github.com/mapbox/frontend-documentation/pull/17#discussion_r202549845). Do you want to take another look &mdash; or maybe @danswick could take a peek?

davidtheclark:
👍 turning over to @danswick

danswick:
This looks great to me! I like the change to a general `card` component and a single `card-container` component. This seems ready to go to me. I just made one small change to the sample gifs to ensure our Firefox friends can view them 😄

PR #27, opened 2018-07-17
Prepare for initial release

danswick:
Adds [build script from mr-ui](https://github.com/mapbox/mr-ui/blob/master/scripts/build-package.js), updates readme, and prepares for `dr-ui` name change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # @mapbox/dr-ui
 
-Pronounced "Doctor UI". **D**ocumentation **R**eact **UI** components. See [@mapbox/dr-ui](https://github.com/mapbox/dr-ui).
+Pronounced "Doctor UI". **D**ocumentation **R**eact **UI** components. See [@mapbox/mr-ui](https://github.com/mapbox/mr-ui).
 
 UI components for Mapbox documentation projects.
 
--- a/README.md
+++ b/README.md
@@ -1,2 +1,50 @@
-# frontend-documentation
-Frontend tools for Mapbox documentation websites.
+# @mapbox/dr-ui
+
+Pronounced "Doctor UI". **D**ocumentation **R**eact **UI** components. See [@mapbox/dr-ui](https://github.com/mapbox/dr-ui).
+
+UI components for Mapbox documentation projects.
+
+**This project is for internal Mapbox usage.** The code is open source and we appreciate bug reports; but we will only consider feature requests and pull requests from Mapbox developers.
+
+## Installation
+
+```
+npm install @mapbox/dr-ui
+```
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,2 +1,50 @@
-# frontend-documentation
-Frontend tools for Mapbox documentation websites.
+# @mapbox/dr-ui
+
+Pronounced "Doctor UI". **D**ocumentation **R**eact **UI** components. See [@mapbox/dr-ui](https://github.com/mapbox/dr-ui).
```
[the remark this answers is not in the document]
danswick: I did! I must have done that first, then clobbered it with find-and-replace later 😖. 

> Looks good. Is it working well?

It is! However, I think I'm going to need to make a small tweak to account for stuff this project needs in addition to components (data about products, custom CSS, etc).

scripts/build-package.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Copy non-src files that we want in the package.
+function copyFiles() {
+  return cpy(['LICENSE', 'CHANGELOG.md', 'README.md'], outputDir, {
+    cwd: rootDir
+  });
+}
+
+function copyCss() {
```
danswick: @davidtheclark this approach feels a it's maybe a bit clunky to me. Do you think it would be preferable to put `copyCss`, `copyData`, and `copyFiles` into a single function or is it ok to split them out like this?

scripts/build-package.js
```diff
@@ -0,0 +1,85 @@
...
+  makeDir(path.resolve(rootDir, 'pkg/css')).then(path => {
+    return cpy(['src/css/*.css'], path, {
+      cwd: rootDir
+    });
+  });
+}
+
+function copyData() {
+  makeDir(path.resolve(rootDir, 'pkg/data')).then(path => {
```
davidtheclark: Typically you'll want to `return` the Promise in a situation like this. That way any errors will be passed through the Promise chain.

scripts/build-package.js
danswick: I tried a couple of different approaches and ran into some issues here, mostly stemming from the fact that i'd like there to be `css` and `data` directories in the finished `pkg` directory: 

- afaict, you can't specify multiple destination directories using `cpy`
- you _can_ use `cpy`'s `--parents` option to preserve directory structure, but we need to call `copyFiles()` outside of the `src` directory (see next bullet), which results in the the copied css and data files being placed into `pkg/src/css/...`, which I don't think makes sense. 
- calling `copyFiles()` from inside the `src` directory didn't work, either. I'm unsure if this is a limitation of `cpy` or something lower level, but using relative paths to parent directories with `cpy` (ex: `../README.md`) doesn't seem to work. 

As a compromise, I feel ok about combining the `copyCss()` and `copyData()` functions (see https://github.com/mapbox/dr-ui/pull/27/commits/1acb9fcee5a53b8a2040eeaa96aab23260488b03), but keeping a separate `copyFiles()` function for all of the items in the root directory. 

What do you think, @davidtheclark?

PR #28, opened 2018-07-17
Remove product menu class

danswick:
After chatting with @davidtheclark, it seems that the best way to handle the custom class in this component is by inlining the style we need instead of relying on a class from some external CSS. Thankfully, `react-popover` has a handy `contentElementAttributes` we can use to do just that! 

It looks like `react-popover` also sets a default `backgroundColor` and setting your own `contentElementAttributes` overwrites that default. I am setting the `backgroundColor` explicitly here and may be missing some other, more crafty way to keep that default value! 

Also noticed that I broke package.json with my last PR, so _this_ PR fixes it 😓

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2644,11 +2644,6 @@
       "integrity": "sha1-Z29us8OZl8LuGsOpJP1hJHSPV40=",
       "dev": true
     },
-    "core-js": {
-      "version": "1.2.7",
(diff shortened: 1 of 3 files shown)


danswick:
@colleenmcginnis smart! that works much better. Went ahead and switched to the assembly class instead.

PR #49, opened 2018-08-10
Adds lil booklet, contact, and troubleshoot images

danswick:
Variety is the spice of life. These little images spice up dr-ui's selection of cute SVGs. 

I've also added a `height` and `width` prop to all of the images so they can be more easily sized without resorting to dark magic. 

image


danswick:
@colleenmcginnis 🤔that could be good. Thinking it might be good to pass in a list of classes to the component, that way you can still handle responsiveness with CSS.

PR #55, opened 2018-09-04
Add LevelIndicator component

### LINKED ISSUE #53: Card: make it possible to add difficulty level and potentially other things
Per discussion in a PR related to the Mapbox help page, it would be useful to generalize the "difficulty level" behavior shown in the "related content" cards at the bottom of tutorials: 

image

colleenmcginnis:
This PR does two things: 

1️⃣ Adds a new `LevelIndicator` component based on the work done in /help (ty @danswick). 

image

2️⃣ Adds optional `level` and `language` props to `Card` (to be used for /help tutorials).

image

Closes #53

### DIFF
--- a/src/components/card/__tests__/card-test-cases.js
+++ b/src/components/card/__tests__/card-test-cases.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Card from '../card';
+import LevelIndicator from '../../level-indicator/level-indicator';
 
 const testCases = {};
 const noRenderCases = {};
@@ -32,6 +33,18 @@ testCases.noImage = {
     title: 'Title',
     path: 'url',
     description: 'described.'
+  }
+};
+
+testCases.withLevelAndLanguage = {
+  component: Card,
+  description: 'Card with level and language',
+  props: {
+    title: 'This is a full-length title',
+    path: 'url',
+    description: "Here's a reasonable length for a description.",
+    level: <LevelIndicator level={2} />,
+    language: 'JavaScript'
   }
 };
 
--- a/src/components/card/__tests__/card-test-cases.js
+++ b/src/components/card/__tests__/card-test-cases.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import Card from '../card';
-import LevelIndicator from '../../level-indicator/level-indicator';
 
 const testCases = {};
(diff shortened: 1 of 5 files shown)


src/components/card/__tests__/card-test-cases.js
```diff
@@ -35,4 +36,16 @@ testCases.noImage = {
...
 };
 
+testCases.withLevelAndLanguage = {
+  component: Card,
+  description: 'Card with level and language',
+  props: {
+    title: 'This is a full-length title',
+    path: 'url',
+    description: "Here's a reasonable length for a description.",
+    level: <LevelIndicator level={2} />,
```
davidtheclark: This is one possibility for the interface; another would be to pass the level number directly. Curious why you chose to pass a React element instead of the level number? Do you think we may want to pass some other element in the future?

src/components/level-indicator/level-indicator.js
```diff
@@ -0,0 +1,65 @@
...
+      borderRadius: '1px'
+    };
+
+    // Make the "difficulty level" squares
+    let levelSquares = [];
+    [1, 2, 3].forEach((num, i) => {
+      let squareColor = '';
+      num > this.props.level
+        ? (squareColor = 'gray-light')
+        : (squareColor = levelColor);
```
davidtheclark: This could be combined into one expression:

```js
const squareColor = num > props.level ? 'gray-light' : levelColor;
```

src/components/level-indicator/level-indicator.js
```diff
@@ -0,0 +1,65 @@
...
+      levelSquares.push(
+        <div
+          key={i}
+          style={squareStyles}
+          className={`inline-block w6 bg-${squareColor} align-middle mr3`}
+        />
+      );
+    });
```
davidtheclark: This is a great use case for `.map()`! When you're taking one array (which should probably be `Object.keys(levels)`?) and transforming every item into another corresponding item in a new array (e.g. a React element), `.map()` is what you're looking for.

src/components/card/__tests__/card-test-cases.js
colleenmcginnis: Yea it's probably safe to just pass the level number.

PR #60, opened 2018-11-08
Add filter option to SectionedNavigation

colleenmcginnis:
Adds the option to add a filter bar that allows you to filter items listed. This attempts to recreate what is done in the Mapbox GL JS docs ([code](https://github.com/mapbox/mapbox-gl-js/blob/master/docs/components/example.js#L87-L104) / [see it in action](https://www.mapbox.com/mapbox-gl-js/examples/)). 

@davidtheclark very open to suggestions if you have any. 

i-did-it

### DIFF
--- a/src/components/sectioned-navigation/sectioned-navigation.js
+++ b/src/components/sectioned-navigation/sectioned-navigation.js
@@ -93,7 +93,8 @@ SectionedNavigation.propTypes = {
 };
 
 SectionedNavigation.defaultProps = {
-  includeCount: true
+  includeCount: true,
+  includeFilterBar: false
 };
 
 export default SectionedNavigation;
--- a/src/components/sectioned-navigation/sectioned-navigation.js
+++ b/src/components/sectioned-navigation/sectioned-navigation.js
@@ -36,9 +36,9 @@ class SectionedNavigation extends React.Component {
             this.setState({ filter: e.target.value }, this.filterResults)
           }
           type="text"
-          className="px6 py6 w-full round border border--gray-light"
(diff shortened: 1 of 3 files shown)


src/components/sectioned-navigation/sectioned-navigation.js
```diff
@@ -16,12 +24,55 @@ class SectionedNavigation extends React.Component {
...
+  renderFilterBar() {
+    const { props } = this;
+    if (!props.includeFilterBar) {
+      return null;
+    }
+    return (
+      <div className="wmax360 mb24">
+        <input
+          onChange={e =>
+            this.setState({ filter: e.target.value }, this.filterResults())
```
davidtheclark: I think you'll want `this.filterResults` instead of `this.filterResults()`. The difference is that `this.filterResults()` would be called as soon as `onChange` is called (because of the `()`), but `this.filterResults` would be referencing the function that would then be called as the callback to `this.setState`.

src/components/sectioned-navigation/sectioned-navigation.js
```diff
@@ -16,12 +24,55 @@ class SectionedNavigation extends React.Component {
...
+      .map(filteredSection => {
+        const filteredItems = filteredSection.items.filter(item => {
+          return item.text.toLowerCase().indexOf(filter) > -1;
+        });
+        return {
+          title: filteredSection.title,
+          url: filteredSection.url,
+          items: filteredItems
+        };
+      });
```
davidtheclark: Nice attention to the details here, keeping sections open and filtering their items.

PR #63, opened 2018-11-12
Little bug fixes

### LINKED ISSUE #57: Icon not centered in BackToTopButton
The up arrow icon is no longer centered in `BackToTopButton`. 

image

cc @langsmith

### LINKED ISSUE #61: Unstick sidebar content when going to a hash
When going directly to a hash very far down on a page ([for example](https://www.mapbox.com/android-docs/maps/overview/styling-map/#style-related-events)) the sidebar content in the `PageLayout` component should not stick, but it does. 

image

@davidtheclark I'm wondering if it's a matter of the order in which things are happening because as soon as I scroll on the page the `Sticky` component snaps up to where it should be. Is there something I could do in componentDidMount if there is a hash in the URL? Or maybe it's more of an issue with the `Sticky` component we're using? Let me know if you have any thoughts that I could further 🕵️‍♀️ .

https://github.com/mapbox/dr-ui/blob/2e874553edb374e68988688c3c6c6ab6578459f4/src/components/page-layout/page-layout.js#L31-L34

cc @samfader

colleenmcginnis:
https://github.com/mapbox/dr-ui/commit/72b9f9abb9b5bd3c291b5528a12c473252e8507e fixes #61 

delay-sticky

https://github.com/mapbox/dr-ui/commit/39eb8dc943eab1a4f381a898013b5ec13aaa492f fixes #57 

screen shot 2018-11-12 at 11 36 45 am

### DIFF
--- a/src/components/page-layout/page-layout.js
+++ b/src/components/page-layout/page-layout.js
@@ -8,7 +8,8 @@ class PageLayout extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      bottomBoundaryValue: 0
+      bottomBoundaryValue: 0,
+      enabled: false
     };
     this.debounceHandleWindowResize = debounce(() => {
       const width = document.body.clientWidth;
@@ -30,6 +31,9 @@ class PageLayout extends React.Component {
 
   componentDidMount() {
     this.debounceHandleWindowResize();
+    setTimeout(() => {
+      this.setState({ enabled: true });
+    }, 500);
     window.addEventListener('resize', this.debounceHandleWindowResize);
   }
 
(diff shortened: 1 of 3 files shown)


src/components/page-layout/page-layout.js
```diff
@@ -8,7 +8,8 @@ class PageLayout extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      bottomBoundaryValue: 0
+      bottomBoundaryValue: 0,
+      enabled: false
```
danswick: Might be worth giving this a more descriptive name for your future debugging self?

colleenmcginnis: Ohhh yes yes good idea.

PR #66, opened 2018-12-06
[NavigationAccordion] collapses subitems when not selected and adds active state

katydecorah:
This PR builds on the work in https://github.com/mapbox/dr-ui/pull/58 and collapses h3 subitems on load. When that section (h2) or a subitem (h3) within the group is selected, then the subitems expand. This PR also adds an active state `txt-bold` to currently selected items in the navigation to help add context.

Unselected, h3s collapsed:
image

Selected h2, h3s expanded:
image

Selected h3, all h3s in section expand:
image

refs  https://github.com/mapbox/api-documentation/pull/613

@danswick to add this functionality, the component requires `props.location.hash` - I'm not sure if this is the best way to pull that info in. Would you happen to have suggestions for a better implementation?


danswick:
@katydecorah I am not very well-versed in getting hashes into props, so I'll just ask some questions and maybe we can figure out a good next step!

In production, would you expect that a router would supply the `location.hash` prop? And would that change what gets passed to `onDropdownChange`? Or maybe the hash changes could all be handled in whatever the `onDropdownChange` function is and the shared accordion component would be a bit more static?

katydecorah:
Per chat, I decided to store the hash location to the state instead

katydecorah:
@HeyStenson or @colleenmcginnis I'd love to get extra eyes on this PR if either of you has time  🙏 Hoping to merge this sometime on Monday so we can do a release

colleenmcginnis:
I can take a look this afternoon @katydecorah!!

colleenmcginnis:
>When that section (h2) or a subitem (h3) within the group is selected, then the subitems expand.

Are you thinking of adding that the h3 headings would only be displayed if the user clicks on an h2 or h3 heading or are you planning to work in updating the hash similar to how https://www.mapbox.com/api-documentation/ currently works so h3s are displayed on scroll, too?

If yes, maybe we should just pass the `NavigationAccordion` an `activeSection` prop that we get using in either (1) the `PageShell` in each docs repo or (2) create a new shared component here that uses both `PageLayout` and `NavigationAccordion` and prescribes what the relationship between these components should be (somewhat similar to the purpose of the `ExamplesPage` component) and adds a scroll event listener to determine the `activeSection` that gets passed to `NavigationAccordion`. 

What do you think @katydecorah?

katydecorah:
@colleenmcginnis that's a really good point, the sidebar navigation should have a scrollspy-like interaction so that as you are scrolling the navigation opens and highlights the current item. I'll work on adding that in!

katydecorah:
In examining how /api-documentation currently handles its scrollspy: each section is wrapped in `div.section`. The site then uses these sections to calculate if a user has scrolled within its boundaries. As far as I can tell, wrapping sections is the best method. I've played around with libraries like https://github.com/makotot/react-scrollspy, which works until you scroll out of the very close vicinity of the heading and results in lots of flashes of state change on the nav.

I think the next best step is to find/create a plugin that we can use in our batfish config that will wrap a div around the start of an h2 and then close it before the start of the next (see example output below). We can then use a very similar implementation as the current /api-documentation to handle the active section and consider also using https://github.com/makotot/react-scrollspy to help highlight the h3s in the nav.

---

Example of desired output:

```markdown
<div class="section">
  <h2>Heading</h2>
  <p>Paragraph</p>
  <h3>Sub heading</h3>
  <p>Paragraph</p>
</div>
<div class="section">
  <h2>Heading</h2>
  <p>Paragraph</p>
  <h3>Sub heading</h3>
  <p>Paragraph</p>
</div>
```

katydecorah:
💋👌 perfect, thanks @colleenmcginnis!

katydecorah:
😅 This is ready for review @colleenmcginnis @HeyStenson ! 

If it's helpful: in /api-documentation I created a testing branch `af-navacc-test` that is using this component.

Notes:
* Created a new plugin that (when included in your batfish.config) will wrap `div.section` around heading levels 2 and 3.
* Added .eslintrc.json to the plugins folder since it required different rules in order to be properly linted

colleenmcginnis:
>If it's helpful: in /api-documentation I created a testing branch af-navacc-test that is using this component.

Awesome! Thanks for setting that up -- taking a look now. 😄 Next quarter maybe we can think about how we can adjust how this repo is set up to meet our testing needs!

package.json
```diff
@@ -66,6 +66,8 @@
     "classnames": "^2.2.6",
     "debounce": "^1.2.0",
     "extend": "^3.0.2",
-    "react-stickynode": "^2.1.0"
+    "lodash.debounce": "^4.0.8",
```
colleenmcginnis: Looks like we're using two different `debounce`s. Maybe we should pick one or the other?

PR #69, opened 2018-12-07
[SectionedNavigation] adds `hideSubItems` option

### LINKED ISSUE #65: Recommendations for SectionedNavgation
While examining the `SectionedNavgation` component I noticed a few patterns that may be confusing:

image

1. The navbar's links have two different outcomes: the headers link within the page while the following links open a new page. I would expect that all links stay on the page or all links go to a new page.
2. Since the navbar duplicates information within the content section, it's almost like we have two views competing on the page. 
3. In some instances the side navbar is long and requires a scrollbar. Users are potentially scrolling twice to navigate similar information.

Recommendation:
* Only show the header links in the navbar and have them only link within the page. Similar to:

image

but with design updates to match the dr-ui theme.

@colleenmcginnis would love your opinion as this component is used by several other doc pages.

katydecorah:
* adds `hideSubItems` option, when this is set to true only the heading for each section will appear (see hidden subitems example below)
* updates design to use similar patterns as other dr-ui elements
* replace input for filter with mr-ui `ControlText` 

subitems | hidden subitems | filter input
------|----|-----
image | image | image

fixes #65


katydecorah:
@colleenmcginnis thanks for catching that! I updated the styles here's what we have now:

image

PR #71, opened 2018-12-07
[Card/CardContainer] styles updates

katydecorah:
This PR makes a few subtle padding and style adjustments to `Card` and `CardContainer`. 

Right now each card has padding that is revealed when hovered by the addition of a shadow. Unfortunately, this adds unused padding to the design and knocks the cards out of alignment with each heading:

image

The changes in this PR, removes the padding around each card and changes the hover interactive to a changing the card's hover text color blue:

image

~~I'd also like to make the images have rounded corners, but I suspect I'll need to PR the ability to pass props in https://github.com/mapbox/appropriate-images~~ Nevermind! We can pass `style` to the element!

### DIFF
--- a/src/components/card-container/card-container.js
+++ b/src/components/card-container/card-container.js
@@ -6,7 +6,7 @@ class CardContainer extends React.PureComponent {
   render() {
     const { props } = this;
     const categoryID = props.path.split('#')[1];
-    const cardClasses = classnames('pb12 col col--12', {
+    const cardClasses = classnames('col col--12 mb36', {
       'col--6-mm': !props.fullWidthCards
     });
     const renderedCards = props.cards.map((card, index) => {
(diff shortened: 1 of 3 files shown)


src/components/card-container/card-container.js
```diff
@@ -6,8 +6,8 @@ class CardContainer extends React.PureComponent {
   render() {
     const { props } = this;
     const categoryID = props.path.split('#')[1];
-    const cardClasses = classnames('pb12 col col--12', {
-      'col--6-mm': !props.fullWidthCards
+    const cardClasses = classnames('col col--12 mb36', {
+      'col--6-ml': !props.fullWidthCards
```
colleenmcginnis: Do you think we need to adjust the width of the images we're using in the cards if we're breaking at `ml` instead of `mm`?

katydecorah: @colleenmcginnis from what I can tell from my testing, I don't believe we'll need to adjust the width of the images. Since the images are being rendered as background images with a `background-size: 100%` they should continue to stretch to fit the container

katydecorah:
> we should make sure we bump dr-ui versions in all docs repos at the same time to make sure this pattern is consistent across all examples pages

@colleenmcginnis that's a really good point, I'll add that as a to do on my end

src/components/card-container/card-container.js
colleenmcginnis: Oh! Yes. I was thinkin' if the widest the div can be is going to be wider than it is now, then we might want to bump the `width` we're using in places like [here](https://github.com/mapbox/android-docs/blob/mb-pages/conf/image-config.js#L48). We can wait and see if the images look pixely. 😝

katydecorah: Aha! I *think* we should be ok considering most pages that use this component also have a sidebar which takes up a lot of real estate. It's unlikely that an image will stretch beyond 800px, but let's keep an eye on it.

PR #83, opened 2018-12-21
adds travis

### LINKED ISSUE #24: Add automated testing
This project does not currently have any automated testing, so all components must be tested manually before publishing. That's probably okay for now while our component count is so low, but will become cumbersome soon. We should add some automated testing once our initial group of components has been added.

katydecorah:
fixes #24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: node_js
+node_js:
+  - '6'
+sudo: false
+notifications:
+  email: false
+cache:
+  directories:
+    - node_modules
+git:
+  depth: 3
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # @mapbox/dr-ui
 
+[![Build Status](https://travis-ci.com/mapbox/dr-ui.svg?branch=master)](https://travis-ci.com/mapbox/dr-ui)
+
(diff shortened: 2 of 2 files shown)


katydecorah:
Now I remember why I stalled on this PR, there's a private NPM module (mapbox-assembly). Looking into it!

katydecorah:
🤦‍♂ apparently the private module `@mapbox/mapbox-assembly` is _only_ used to run the test cases app. I moved the module as an optional dependencies so that Travis can run the tests.

package.json
```diff
@@ -72,5 +71,8 @@
     "react-html-parser": "^2.0.2",
     "react-stickynode": "^2.1.1",
     "rehype-sectionize-headings": "^1.0.0-rc.1"
+  },
+  "optionalDependencies": {
+    "@mapbox/mapbox-assembly": "^0.26.0"
```
colleenmcginnis: Oh! I think we can update to `"@mapbox/mbx-assembly": "^0.28.0"` (or higher), which is when we switched the mapbox assembly package to public. https://github.com/mapbox/mbx-assembly/blob/master/mapbox-changelog.md#0280

katydecorah: 🤦‍♂ thank u colleen

PR #95, opened 2019-01-19
catches up with mr-ui

katydecorah:
I still need to double check that there weren't any breaking changes with some of these other mr-ui components. So far only one component was deprecated and this PR reworks it.

fixes #71


katydecorah:
I went through mr-ui's changelog and checked for breaking changes against the components that we're using here and made one more small change.

PR #96, opened 2019-01-23
[PageLayout] adds optional prop to change the column size of the sidebar

katydecorah:
The width of the sidebar and content is fixed in dr-ui's `PageLayout`. Sometimes our content asks for more flexibility. This PR adds an optional prop `sideBarColSize`, a number. If available, dr-ui will use that number for the sidebar column: `col--[sideBarColSize]-ml` and then compute the content column size: `col--[12 - sideBarColSize]-ml` size.

We'll keep the `-mm` col sizes as they are to make sure that content doesn't get too squished at that breakpoint.

For quality control, sideBarColSize must be > 2 and < 7. Otherwise it will ignore the value and use the default column sizing.

### DIFF
--- a/src/components/page-layout/__tests__/page-layout-test-cases.js
+++ b/src/components/page-layout/__tests__/page-layout-test-cases.js
@@ -17,7 +17,7 @@ testCases.basic = {
 };
 
 testCases.smallSidebarCol = {
-  description: 'Basic, small sideBarCol',
+  description: 'Custom sidebar column size',
   component: PageLayout,
   props: {
     sidebarContent: <div>Some content</div>,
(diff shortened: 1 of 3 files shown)


src/components/page-layout/__tests__/page-layout-test-cases.js
```diff
@@ -16,6 +16,47 @@ testCases.basic = {
...
+    sidebarContentStickyTop: 0,
+    sidebarContentStickyTopNarrow: 0,
+    sideBarColSize: 3,
+    children: <div>Doc content</div>
+  }
+};
+
+testCases.tooSmallSidebarCol = {
+  description:
+    'Custom sidebar column size, too small - default to original size',
```
colleenmcginnis: Wow this is brill.

PR #97, opened 2019-01-23
[PageLayout] adds optional interactiveClass prop to watch for changes on click

katydecorah:
On the new mapbox-gl-js page, when you start expanding the api items, the length of the page changes. However, since we aren't listening for changes to the page height the sticky sidebar gets unmounted from its position too early: 

image

This PR adds an optional prop that you can pass a class that you want to watch for clicks. If that element is clicked then the height of the page will be recalculated and the sticky sidebar will adjust and stay mounted until the bottom of the page.

@colleenmcginnis for review 🙇‍♀️ this is a little tricky to test in our current testing suite


katydecorah:
@colleenmcginnis sounds like a good plan! I added comments to document  this prop and the previous one (`sideBarColSize`) #96

PR #102, opened 2019-01-29
Add warning and error themes to `Note`

### LINKED ISSUE #94: Add themes to `Note`
Theme ideas:

* error
* warning

I noticed android-docs has a WarningNote:

image

katydecorah:
@colleenmcginnis do you know offhand if we have any sites using the Note component and passing a custom theme to it? This PR proposes to remove passing style-based themes and replace it with keyword themes (default/no theme specified, warning, error). If it's important to be able to customize the note, I can adjust!

fixes #94

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 - Add prism.css to create shared syntax styles [#108](https://github.com/mapbox/dr-ui/pull/108)
 - Add new styles for fullwidth `Card` and `CardContainer`. [#104](https://github.com/mapbox/dr-ui/pull/104)
 - Add CSS to set max-height and style scrollbars for prose code examples. [#103](https://github.com/mapbox/dr-ui/pull/103)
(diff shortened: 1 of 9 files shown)


colleenmcginnis:
I can't think of any sites where a custom `Note` is added directly to an md file. The only instance I know of is the `WarningNote` component https://github.com/mapbox/android-docs/blob/publisher-production/src/components/warning-note.js#L23-L33 (which I know you're already aware of 😉).

katydecorah:
@colleenmcginnis I decided to reduce the number of example notes. I couldn't think of a good example for a basic not where we'd need to change the image to something different. Please let me know if you have any use cases!

image

image


I've also updated the props for all images to replace `width` and `height` with `size` since we always want those images to be round.

Ready for review!

katydecorah:
Thanks for reminding me about Floater, I just a ticket stub for that one! https://github.com/mapbox/dr-ui/issues/109

PR #104, opened 2019-01-29
[Card/CardContainer] Add new styles for fullwidth cards

### LINKED ISSUE #92: [Card] apply different styling when no image
If a `Card` has no image, the text-based layout produces quite a bit of spacing. I'm thinking we could switch the style a little bit to adjust the spacing and add a border divider, something like this:

image

katydecorah:
For extra testing, I created a branch in help called `DNM-cardcontainer-example` that uses these components.

fixes #92 

## Before

image

## After

image

### DIFF
--- a/src/components/examples-page/__tests__/examples-page-test-cases.js
+++ b/src/components/examples-page/__tests__/examples-page-test-cases.js
@@ -23,7 +23,7 @@ testCases.basic = {
             key="0"
             title="Example one"
             path="path"
-            description="description"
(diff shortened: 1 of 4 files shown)


src/components/examples-page/__tests__/examples-page-test-cases.js
```diff
@@ -117,4 +117,61 @@ testCases.basic = {
   }
 };
 
+testCases.noThumbs = {
```
colleenmcginnis: 🙅 👍

PR #115, opened 2019-02-26
Add optional prop for tags for first level items in NavigationAccordion

colleenmcginnis:
Related: https://github.com/mapbox/android-docs/pull/842

Adds an additional prop to `NavigationAccordion` for adding a tag to a first level item.

screen shot 2019-02-25 at 9 01 00 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Master
+
+- Add an optional prop in `NavigationAccordion` to add tags to first level items. (#115)(https://github.com/mapbox/dr-ui/pull/115)
+
 ## 0.6.0
 
 - Add ability to filter the description if it's available in `SectionedNavigation`. (#111)(https://github.com/mapbox/dr-ui/pull/111)
--- a/src/components/navigation-accordion/__tests__/navigation-accordion-test-cases.js
+++ b/src/components/navigation-accordion/__tests__/navigation-accordion-test-cases.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import NavigationAccordion from '../navigation-accordion';
 
 const testCases = {};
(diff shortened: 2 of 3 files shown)


katydecorah:
@colleenmcginnis I think we should create components for these tags as well to make sure they remain consistent. I can lend a hand, if needed!

We'll also want to make sure the color contrast passes since smaller text requires a higher contrast ratio.

### Advanced
- Updating to `color-blue-dark bg-blue-faint` is sufficient. 
- However in `level-indicator` advanced is red. Should we be consistent with that component? If so, updating to `color-red-dark bg-red-faint` is sufficient.

### Fundamentals

- Updating to `color-pink-dark bg-pink-faint` is sufficient.
- What do you think about using green instead? We use green in `level-indicator` for "beginner". Updating to `bg-green-faint` and `style={{color: "#1c7d50"}}` would get us there.

colleenmcginnis:
>I think we should create components for these tags as well to make sure they remain consistent.

Yes! However, I was holding off for two reasons: 
- This is the first set of docs where we're trying out this system. Right now https://github.com/mapbox/android-docs/pull/842 is only using one tag ("fundamentals"). I'm not sure what other tags we might want to use. I'd like to wait until we start adding this to other docs. 
- I think we might be able to collaborate with @lshig on https://github.com/mapbox/mr-ui/issues/72. 👯

colleenmcginnis:
>* However in `level-indicator` advanced is red. Should we be consistent with that component? If so, updating to `color-red-dark bg-red-faint` is sufficient.

The `Advanced` tag was totally hypothetical -- there's currently no use for it. I shouldn't have used a term that conflicted with the level-indicator tag. 🙈 

>* What do you think about using green instead? We use green in `level-indicator` for "beginner". Updating to `bg-green-faint` and `style={{color: "#1c7d50"}}` would get us there.

The guides that I'm hoping to tag as `fundamentals` in https://github.com/mapbox/android-docs/pull/842 aren't really the equivalent of `beginner`. The purpose of the tag is more like... start here because 🗣 THIS IS IMPORTANT rather than because it's the simplest. I'm open to other words to replace `fundamentals` if you have any suggestions!

PR #117, opened 2019-03-05
create versionSort function

katydecorah:
This PR creates a helper function to sort version numbers for the API reference dropdowns


katydecorah:
@colleenmcginnis 

* When creating tests for the helper function, the console insisted that I update the babel versions so that's why you'll see some updates to babel and using the namespaced versions.
* I copied the version-sort.js file and pasted it (temporarily) into the iOS PR so we can test this function. Once we feel good about it, then we can do a release and update the iOS PR

colleenmcginnis:
@katydecorah I'm getting this error after running `npm ci` and then `npm start` 🤔 :

```
module.js:557
    throw err;
    ^

Error: Cannot find module 'babel-core'
...
```

katydecorah:
@colleenmcginnis whoops! I pushed a fix to update babelify to fix that error 🙈

PR #121, opened 2019-04-10
Create Search component

katydecorah:
This PR works to add a new Search component that will add an input hooked up to Swiftype. Results are displayed in a popover under the input.

To test:
1. npm start
2. Navigate to http://192.168.7.223:9966/Search

The test has two examples - basic search and in context of where I think it could go on each site (on the right-end of the top bar sticker).

# QA Checklist 

## Responsive

- [x] Check on mobile-size viewport.
- [x] Check on tablet-size viewport.
- [x] Check on laptop-size viewport.
- [x] Check on big-monitor-size viewport.

## Cross-browser

- [x] Check in Chrome.
- [x] Check in Firefox.
- [x] Check in Safari.
- [x] Check in Edge.
- [ ] Check in IE11.
- [x] Check in mobile Safari.
- [x] Check in Android's Chrome.


katydecorah:
@colleenmcginnis this is ready for review! 

Note: search doesn't appear at all in IE11: 

image

I haven't pinpointed the exact point of failure, but I feel ok with punting on getting the search to work in IE11 since it gracefully degrades (to no search) and doesn't break the layout.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Master
+
+- Add `Search` component
+- 🚨 [Breaking change] Remove `txt-bold` and `txt-s` from `LevelIndicator` component to make it more flexible.
```
katydecorah: When we update dr-ui in other repos using `LevelIndicator` we'll want to wrap the component in a span or div with these classes where needed.

src/components/search/search-box.js
```diff
@@ -0,0 +1,90 @@
...
+                          key={index}
+                          result={result}
+                          index={index}
+                          downshiftProps={downshiftProps}
+                        />
+                      ))}
+                    </ul>
+                  ) : (
+                    <div className="py12 px12">
+                      Sorry, we didn't find anything.
```
katydecorah: Copy suggestions welcome!

src/components/search/search-box.js
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Downshift from 'downshift';
```
katydecorah: 💡Downshift makes our search input and and results navigable by keyboard!

src/components/search/search-result.js
```diff
@@ -0,0 +1,109 @@
...
+                  <span>
+                    {site}
+                    <Icon name="chevron-right" inline={true} />
+                  </span>
+                ) : (
+                  ''
+                )}
+                {subsite && subsite !== title && subsite !== site ? (
+                  <span>
+                    {subsite.replace(/\sfor\s(iOS|Android|Vision)/, '')}
```
katydecorah: 💡This temporarily removes the "for Android" (or iOS or Vision) from the subsite name. We should eventually update the docs-page-shell to strip this out but this will fix it on the frontend to give us cleaner looking results.

src/components/search/search-box.js
colleenmcginnis: ooooooo!

src/components/search/search-box.js
colleenmcginnis: 🤔 @HeyStenson any thoughts here?

src/components/search/search-result.js
colleenmcginnis: Good call! Much cleaner. Should we do the same for Unity? 

image

src/components/search/search-box.js
HeyStenson: Maybe something like: `Hmmm, we didn't find anything. Reword your search, or contact Support.` (with a link to the support page). 

❓❓❓

katydecorah: I love that @HeyStenson! I'll add it 🙇‍♀

PR #130, opened 2019-05-15
[search] add "collapse" option

katydecorah:
If `collapse={true}` then the search component will collapse into a button:

> image

When the button is clicked/focused, the input will expand. The input will overlap anything to its left so that it will fit in the space without breaking the layout:

> image

Once the user clicks way from the opened input, it will collapse again.

On smaller screens/mobile, the collapse option disables to reveal the full input:

> image

# QA

## Responsive

- [x] Check on mobile-size viewport.
- [x] Check on tablet-size viewport.
- [x] Check on laptop-size viewport.
- [x] Check on big-monitor-size viewport.

## Cross-browser

- [x] Check in Chrome.
- [x] Check in Firefox.
- [x] Check in Safari.
- [x] Check in Edge.
- [ ] Check in IE11.
- [x] Check in mobile Safari.
- [x] Check in Android's Chrome.

# Review

⏲ No rush

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10213,6 +10213,18 @@
       "resolved": "https://registry.npmjs.org/react-submittable/-/react-submittable-2.0.0.tgz",
       "integrity": "sha512-xenO5Z0Ibrvv0CxQwFOj+nyxim/Nznwbv5cG+tp6e9+ZvS5Ku518HM+BxfTtySc0RbKnmSSJjuydoY/8odDbIA=="
     },
+    "react-test-renderer": {
+      "version": "16.8.6",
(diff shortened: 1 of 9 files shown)


katydecorah:
@colleenmcginnis No rush on this review since we likely won't roll side-wide search out until after next week

colleenmcginnis:
@katydecorah the code looks great! The search bar popping over the tab text was not sitting right with me so I started playing around with another approach (which def would need to be refined) in https://github.com/mapbox/dr-ui/pull/133. Let me know what you think!

katydecorah:
closes in favor of #133

PR #131, opened 2019-05-17
Add EnterpriseFlag, add option to ProductMenu and OveriewHeader

colleenmcginnis:
Adds an Enterprise flag for products and features that are only available for users with Enterprise accounts. Then adds the option to add the flag to the ProductMenu and OverviewHeader components similar to the BetaFlag component. 

This works well for the OverviewHeader component, but the length of the word `Enterprise` doesn't work so well with the ProductMenu (especially for a product that might be both in Beta and for Enterprise accounts only). @katydecorah do you have any ideas here? 

Screen Shot 2019-05-16 at 5 38 40 PM

Screen Shot 2019-05-16 at 5 40 16 PM

### DIFF
--- a/src/components/enterprise-flag/__tests__/enterprise-flag-test-cases.js
+++ b/src/components/enterprise-flag/__tests__/enterprise-flag-test-cases.js
@@ -1,0 +1,20 @@
+import EnterpriseFlag from '../enterprise-flag';
+
+const testCases = {};
+const noRenderCases = {};
+
+testCases.basic = {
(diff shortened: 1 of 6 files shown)


katydecorah:
@colleenmcginnis for the ProductMenu: how about we tuck those tags above the text? 

image


If this looks good to you, I can push the code!

colleenmcginnis:
There's just not enough room for a tag like this -- I think we should just indicate that products are Enterprise-only in a note instead (like we already do).

PR #132, opened 2019-05-23
Adjust version-sort for XX.XX.XX format

colleenmcginnis:
`version-sort.js` is not working with numbers that have two digits in the middle number in the `XX.XX.XX` format: 

image

This change _should_ set us up for success if any of the three numbers are 1 or 2 digits by prepending single-digit numbers with `0` before building the `num` value.

@katydecorah I can't get one test to pass. I would also love your thoughts on this approach! 

cc @friedbunny @captainbarbosa @riastrad


katydecorah:
@colleenmcginnis I ended up swapping out our sorting function for an npm module that will do that logic for us. I think we got it 😅

colleenmcginnis:
Works beautifully 💅

PR #134, opened 2019-05-30
increase zIndex for sticky sidebar

katydecorah:
While reviewing https://github.com/mapbox/studio-manual/pull/73 I noticed that the sticky sidebar has a zIndex fight with the illustrations.

image

We updated the zIndex for the TopBarSticker to 3 in https://github.com/mapbox/dr-ui/pull/67 and I think we should do the same for the sidebar sticky navigation.


katydecorah:
🤦‍♂ thanks for catching that. I think the best method is to bump up TopbarSticker

PR #135, opened 2019-05-30
Feedback component

katydecorah:
ref https://github.com/mapbox/documentation/issues/176

This PR sketches out the `Feedback` component that will appear on the bottom of pages and sections:

> image

User selects "Yes" | User selects "No"
---|---
image | image

## Tracking response

* On Yes/No button click the response will be tracked in Segment.
* Text feedback will be sent to Zendesk via the Ticket API. We'll also track in Segment if the user submits text feedback.

## Options

* The `type` prop (possible values: `page` or `section`; default is `page`) allows use to switch out the context for the feedback text:
> image
* The `site` prop (required) allows us to track the current site in Segment for context in our dashboards.
* The `section` prop (optional) allows us to track the section of a given page in Segment. This will be handy for long pages with multiple feedback components.

## To do

* [x] Review/revise all copy
* [x] Identify if there are any other feedback "types" to include other than page and section.
* [x] What does it look like after the users submits text feedback?
* [x] Hook up Segment tracking (yes, no, submit feedback)
* [x] Hook up Zendesk to post feedback with the [Ticket API](https://developer.zendesk.com/rest_api/docs/support/tickets#create-ticket)
* [x] Run through QA checklist

---

@HeyStenson @colleenmcginnis


katydecorah:
Instead of using the forward-event client via the private npm module, I [copied over the client](https://github.com/mapbox/forward-event/blob/master/forward-event-client/index.js) as a new function `src/components/feedback/forward-event.js`. In each repo we'll save the webhook URLs (staging and production) in constants.json and then pass them to the Feedback component. 


### Reviewer checklist

1. `npm install && npm start` 
2. Click yes/no on the feedback module
3. Open [segment debugger](https://app.segment.com/mapbox/sources/mapbox-staging/debugger) and search for "sent docs feedback" to see if your events came through using the following testing environments:

* [ ] Logged in:
  * [ ] Ad blocker is on and the event in segment captured my username.
  * [ ] Ad blocker off and the event in segment captured my username.
  * [ ] No errors logged to console (other than marketo, segment, customerio when adblocker is on)
* [ ] Logged out:
  * [ ] Ad blocker is on and the event in segment did not capture my username but the event is present.
  * [ ] Ad blocker is off and the event in segment did not capture my username  but the event is present.
  * [ ] No errors logged to console  (other than marketo, segment, customerio when adblocker is on)
* [ ] Running `npm run build` returns no errors

Logged out events should look similar to:
```js
analytics.track({
  anonymousId: '5eea70cc...',
  event: 'Sent docs feedback',
  properties: {
    helpful: true,
    page: {},
    site: 'Help'
  }
});
```

Logged in events should look similar to:
```js
analytics.track({
  userId: 'decorah',
  event: 'Sent docs feedback',
  properties: {
    helpful: true,
    page: {},
    site: 'Help',
    userId: 'decorah'
  }
});
```

katydecorah:
This is ready for review. I added a checklist of things to look out for in the comment above.

I also moved this exact component to /help to make sure some of the changes work (passing the username and webhook to the component) and pushed those changes to staging. You can test either /help staging and/or the feedback component in this repo.

katydecorah:
Per chat, I just pushed a commit to improve the confirmation messages to help make it clear that the textarea is optional. I also just pushed this update to /help staging, so that should be live in a few minutes.

HeyStenson:
Capturing this so that we can discuss later -

We may need to make it explicit to users that this is _not_ a way to request support. We could add language like the following, plus a link to https://support.mapbox.com/hc/en-us, as a way to divert users who are seeking specific help:

```
If you have a technical question about how to use a Mapbox product, contact Support.
```

Or even more explicit:

```
This form is for documentation feedback. If you have a technical question about how to use a Mapbox product, contact Support. 
```

In terms of placement, I think this might work best _under_ the text form, but _above_ the "Send feedback" button. 

I'll think more about wording/placement and we can discuss later - just wanted to get these thoughts down while they were fresh!

katydecorah:
@HeyStenson I think your language suggestions are spot on 💅 

Here's the current state:

image

Is that what you had in mind?

HeyStenson:
@katydecorah YES 🙌🙌🙌 This looks awesome!

colleenmcginnis:
🤔  v small detail -- should send feedback button be bigger or maybe be above the note about support?

image

colleenmcginnis:
@katydecorah I'm having a hard time testing -- I'm seeing the events show up in Segment 👍 , but I don't see a username even when logged in 👎.

katydecorah:
> should send feedback button be bigger or maybe be above the note about support?

@colleenmcginnis Yes, this makes sense! I dropped the note to the bottom:

image

PR #137, opened 2019-05-31
a catalog site 

### LINKED ISSUE #23: Add docs build process maybe
We could add docs to this project a-la [mr-ui](https://github.com/mapbox/mr-ui/blob/master/README.md#development). @colleenmcginnis -- do you think that would be useful and/or a good thing to prioritize?

katydecorah:
Borrowing from our very good friend mr-ui

image

To do:
- [x] TopbarSticker and PageLayout travel with the page, need to restrict the sticky to the example.

fixes #23


katydecorah:
In testing the deploy script, the catalog site is now live at: https://mapbox.github.io/dr-ui/

After working with adding examples and testing out the components, there are a few things that I think should be fixed in separate PRs:
* [ ] all images: 
  - set a default `size` instead of making it required
  - compress SVG
  - make `size` value a number
* [ ] CodeToggle: refactor code so that it works as a standalone component.

katydecorah:
Gave this a quick update to get a slew of changes in - the one thing that feels cumbersome is that we essentially duplicate the work that in each test case example in `__tests__/*-test-cases.js`. While it's possible to import the content from a given test-cases file, the code preview shows all that noisy importing of testcases and not the actual component 🤔

katydecorah:
> The one thing that feels cumbersome is that we essentially duplicate the work in each test case (__tests__/*-test-cases.js) to create a examples/*.js file for the site. 

I think a good workflow is to replace some test cases with examples. Here's what that would look like for the back-to-top-button component:

`/back-to-top-button/examples/basic.js`
```jsx
/*
Basic.
*/
import React from 'react';
import BackToTopButton from '../back-to-top-button';

export default class Example extends React.Component {
  render() {
    return <BackToTopButton />;
  }
}
```

`/back-to-top-button/__tests__/back-to-top-button-test-cases.js`
```
import React from 'react';
import BackToTopButton from '../back-to-top-button';
import Basic from '../examples/basic';

const testCases = {};
const noRenderCases = {};

testCases.basic = {
  description: 'Just the button',
  element: (
      <Basic />
    )
};

testCases.tall = {
  description: 'Gave this button some breathing room',
  element: (
    <div style={{ marginTop: '1000px' }}>
      <BackToTopButton />
    </div>
  )
};
```

katydecorah:
This is ready for review. 

The current version is deployed at https://mapbox.github.io/dr-ui/

## Overview of changes

- **New CONTRIBUTING.md.** The catalog site adds a slightly new workflow that I'm attempting to document in the new [CONTRIBUTING.md file](https://github.com/mapbox/dr-ui/blob/catalog/CONTRIBUTING.md). 
  - ❓ I'd love your gut checks on this file if anything is missing or doesn't make sense. This workflow will make sure the catalog and test cases stay up to date and prevent duplicating code where possible.
  - ❓ Are two sites too much to manage? I'm hoping not. I think we can continue to do our development working using the test cases app and won't likely need to run the catalog site during dev work.  I considered doing away with the test cases app, but there are quite a few one off cases that could overwhelm the catalog site.

- **New component script.** This PR includes a generation script for creating new components. If you run `node scripts/new.js crocs image` it will automatically create all the files and starter code for the `CrocsImage` component. (Including creating a basic catalog example and importing it as a test case.).

- **New tests.** There are a few more tests in this repo to help assert catalog data. Travis will also run `npm run build-docs` - this will help us catch build issues of all components. Fixes https://github.com/mapbox/dr-ui/issues/168

- **Ordered components in catalog by category.** I thought it would be helpful to order all the components in the catalog by category (Features, Code, Layout, Images) to improve discoverability. The categories are controlled by docs/src/categories.json and asserted by tests.  
  - ❓ What do you think of these categories? Would you change or move anything?

There should be _no_ changes to the components themselves in this PR. In most cases I used a component's first (basic) test case as the catalog example. I did make some slight changes to a few examples and test cases to make them more palatable for the catalog.

PR #138, opened 2019-06-03
[search] add ability to filter results to current site versus all docs

katydecorah:
If you set the `site` prop, it will add a toggle at the top of the results so that the user can filter results for the current site or across all docs.

Unfortunately, we cannot default the toggle to the current site (waiting for sticky filters upstream in search-ui). Without sticky filters, any filter we set manually gets reset when the user searches. Hopefully we'll be able to add this in the near future.

> image

> image

How to test:

1. npm run start
2. Navigate to http://192.168.7.223:9966/Search
3. Try the second test case: "Basic search with `site` set to show filter toggle"

### DIFF
--- a/src/components/search/search-box.js
+++ b/src/components/search/search-box.js
@@ -4,6 +4,8 @@ import Downshift from 'downshift';
 import SearchModal from './search-modal';
 import SearchResult from './search-result';
 import debounce from 'debounce';
+import { getFilterValueDisplay } from '@elastic/react-search-ui-views/lib/view-helpers';
+import { Facet } from '@elastic/react-search-ui';
 
 class SearchBox extends React.Component {
   constructor(props) {
@@ -16,6 +18,7 @@ class SearchBox extends React.Component {
     this.closeModal = this.closeModal.bind(this);
     this.renderModal = this.renderModal.bind(this);
     this.renderSearchBar = this.renderSearchBar.bind(this);
+    this.singleLinksFacet = this.singleLinksFacet.bind(this);
     this.checkWidth = debounce(() => {
       const width = document.body.clientWidth;
       this.setState({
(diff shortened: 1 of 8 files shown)


katydecorah:
Note: I'm seeing an zIndex fight with the modal and topbar sticker. Going to fix in another PR.

colleenmcginnis:
@katydecorah I'm getting a little lost in this code. Would the filtered option always equal to the part of the url that comes after `docs.mapbox.com/{THIS_PART}`?

katydecorah:
@colleenmcginnis it's filtering based on the value of `site` in Swiftype. We set`site` via the page shell: https://github.com/mapbox/api-documentation/blob/publisher-production/src/components/page-shell.js#L105 which creates the Swiftype meta tag.

src/components/search/search-box.js
```diff
@@ -41,6 +44,55 @@ class SearchBox extends React.Component {
...
+  singleLinksFacet = ({
+    className, // eslint-disable-line
+    label, // eslint-disable-line
+    onRemove,
+    onSelect,
+    options,
+    values = []
+  }) => {
+    const value = values[0];
+    const siteFilter = options.filter(opt => opt.value === this.props.site)[0];
```
katydecorah: @colleenmcginnis to further illustrate:

Swiftype provides a list of unique values stored for the field `site` this function filters to find the `site` prop we send the to Search component. In this case we're looking for the value `API`. Then we can use the data from Switype to create the "API" toggle.

HeyStenson:
This looks awesome @katydecorah! Ability to filter by the site they're on will be a great addition to the user experience.

src/components/search/search-box.js
colleenmcginnis: Ohhh thanks @katydecorah! 💡

katydecorah:
> I think this makes sense, but something to consider is that by filtering results on pages in repos other than /help will prevent results from /help from showing up.

@colleenmcginnis This is a limitation that I don't think we have a solution right now. We could watch the metrics to see if this is a problem (for example if users are searching on iOS and Android and there's an elevated number of search exits and/or no results returned).

PR #140, opened 2019-06-03
[search] adds segment tracking for queries and clicks

katydecorah:
This PR adds tracking to our Search component for:
1. Queries typed
2. Result clicked

How to test:
1. npm install
2. npm start
3. open segment > sources > mapbox-staging (leave window open) 
4. in another window: navigate to http://192.168.7.223:9966/Search
5. type a query, click a result
6. switch back to segment and click "Pause" and search for "Searched docs" events. Click each one. You should see an event for the queries and an event for the clicked item:

image
 
When you click "Raw" you'll see the metadata attached with the event, including page searched from 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "@mapbox/eslint-config-mapbox": "^1.2.1",
     "@mapbox/mbx-assembly": "^0.28.2",
     "@mapbox/react-test-kitchen": "^0.1.3",
+    "@mapbox/web-analytics": "^0.5.1",
     "babel-eslint": "^8.2.6",
     "babelify": "^10.0.0",
     "budo": "^11.6.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "@mapbox/eslint-config-mapbox": "^1.2.1",
     "@mapbox/mbx-assembly": "^0.28.2",
     "@mapbox/react-test-kitchen": "^0.1.3",
+    "@mapbox/web-analytics": "^0.5.1",
```
katydecorah: Note: this is a private module. Including it makes it far easier to test segment events, but not easy for teammates who don't have access. I think we should leave for now for testing purposes, but can circle back if it becomes a blocker for others.

colleenmcginnis: Is this still an issue now that it's moved to optional dependencies?

katydecorah: @colleenmcginnis now that it's an optional dependencies, I don't think non-aws users will have an issue install this repo locally.

Also want to note that web-analytics is installed here for local testing only (so we can make sure events are getting tracked) - it won't get bundled in the dr-ui build.

PR #154, opened 2019-07-15
Create second-level-items.js helper function

katydecorah:
This helper function will build the `secondLevelItems` prop for `NavigationAccordion`.

To test this function, I pulled real data from 4 sites (android, api, vector-tiles, gl-js) and then used the function to make sure it returns the same data that the site currently returns. (This work is shown in the test file in this PR.)

🐌


katydecorah:
I'm going to punt on this

Issue #153, opened 2019-07-15
More meta functions: navigation, headings


katydecorah:
The `page-shell` files in repo have a lot of repeatable code to build page metadata (headings/table of contents) and the site's navigation.

We should build helper functions to help make this code consistent and testable:

* [ ] Meta (frontmatter) #141 
* [ ] Site navigation
* [ ] Headings, table of contents

PR #155, opened 2019-07-15
Create ordered-pages.js

katydecorah:
This helper function will replace the `orderedPages` value found in all batfish.config.js sites.

Refs #153


katydecorah:
I'm going to punt on this

PR #158, opened 2019-07-23
[search] do not load component on IE 11

katydecorah:
We're seeing an issue in IE11 where the /search-ui dependencies are throwing errors due to es2015 features. This PR adds babel polyfills to the search component.

I tested this locally in /mapbox-gl-js and the page and maps can now load without error:
image

The only issue to still solve is that the Search button now appears in IE11, but no results are returned when typing a query.


katydecorah:
image
> IE 11 🎉 


New plan: if we prevent the search component from loading on IE11 then we won't need to use a babel polyfill.

Here's how to test this PR:

1. Pull down this PR, `npm install` then `npm run build`
2. Pull down this GL JS PR, in package.json, replace the dr-ui version with the path to your local dr-ui, for example `"@mapbox/dr-ui": "file:/Users/katydecorah/Documents/GitHub/dr-ui/pkg",`
3. (You may need to delete node_modules first) Run `yarn install`
4. Run `yarn run start-docs`
5. In Browserstack open `http://localhost:8080/mapbox-gl-js/example/simple-map/` or any other example page to confirm the page and map loads. There will be no search bar.

PR #162, opened 2019-08-03
Create a note "flavor" for new products or features

HeyStenson:
@brsbl noted in our update of the Tilesets API docs that:

> we usually use orange to denote things like deprecation/bad things and I wonder if a different color signifying a good thing (light green?) might make more sense?

This makes sense to me! This PR creates a new Note flavor to indicate that _a product or feature is new_.  

Screen Shot 2019-08-02 at 4 59 00 PM

@mapbox/docs to review, please 🙏

### DIFF
--- a/src/components/new-image/new-image.js
+++ b/src/components/new-image/new-image.js
@@ -13,7 +13,7 @@ export default class NewImage extends React.Component {
         }}
       >
         <div className={`bg-${props.color || 'blue'}-light px3 round-full`}>
-          <svg viewBox="0 0 18 18" className={`color-${props.color || 'blue'}`}>
+          <svg viewBox="0 0 18 18" className={`color-${props.color || 'blue'} mt3`}>
             <title>warning</title>
             <path
               style={{ fill: 'currentColor' }}
--- a/src/components/new-image/new-image.js
+++ b/src/components/new-image/new-image.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class NewImage extends React.Component {
+  render() {
+    const { props } = this;
+    return (
+      <div
+        style={{
+          padding: '5px',
+          width: `${props.size}px`,
+          height: `${props.size}px`
+        }}
+      >
+        <div className={`bg-${props.color || 'blue'}-light px3 round-full`}>
+          <svg viewBox="0 0 18 18" className={`color-${props.color || 'blue'}`}>
+            <title>warning</title>
+            <path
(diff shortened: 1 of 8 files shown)


src/components/new-image/new-image.js
```diff
@@ -0,0 +1,32 @@
...
+    return (
+      <div
+        style={{
+          padding: '5px',
+          width: `${props.size}px`,
+          height: `${props.size}px`
+        }}
+      >
+        <div className={`bg-${props.color || 'blue'}-light px3 round-full`}>
+          <svg viewBox="0 0 18 18" className={`color-${props.color || 'blue'}`}>
```
katydecorah: Recommend adding the `mt3` class to this svg element to add 3 pixels to the top to make the icon a little more centered in the circle:

```suggestion
          <svg viewBox="0 0 18 18" className={`color-${props.color || 'blue'} mt3`}>
```

Result:

image

src/components/note/note.js
```diff
@@ -24,6 +24,10 @@ class Note extends React.Component {
       error: {
         background: '#fbe5e5',
         color: '#ba3b3f'
+      },
+      new: {
+        background: '#e8f5ee',
+        color: '#269561'
```
katydecorah: Recommend a slightly darker text color to improve the color contrast: https://webaim.org/resources/contrastchecker/?fcolor=269561&bcolor=E8F5EE

```suggestion
        color: '#1b7d4f'
```

Color contrast with suggestion: https://webaim.org/resources/contrastchecker/?fcolor=1B7D4F&bcolor=E8F5EE

src/components/new-image/new-image.js
HeyStenson: Ooh, yes! Much prettier!

src/components/note/note.js
HeyStenson: Good thinking! I need to get better about remembering to use the contrast checker.

HeyStenson:
Thanks for the styling suggestions @katydecorah - they look great! With dr-ui, is the next step to merge this PR into `master`, then it will get added to the next release?

katydecorah:
> With dr-ui, is the next step to merge this PR into master, then it will get added to the next release

@HeyStenson yes! Recommend updating CHANGELOG.md by adding a new `## Master` heading to the top of the page and then a bullet point explaining this change. This will help keep the changes in master straight.

The tests are failing because we made some slight UI adjustments and the snapshot tests need to be updated, you can do that by running `npm test -- -u` and then committing the files that command generated.

PR #164, opened 2019-08-05
Babel updates: compile helpers, plugins; add plugins

katydecorah:
This PR is part of an action item from the IE11 bug postmortem.

* Makes babel compile `src/helpers` and `src/plugins` files, these were previously copied over. 
   - While rolling out 0.19.3, repos that used the `version-sort` helper were failing on IE11 due to an arrow function. I was able to get around it by adding `dr-ui` to the `babelInclude` in those repos.
   - This PR fixes that at the source by adding the transform arrow functions plugin.
   - We don't have a `src/data` dir, so that is removed fro the build script.
* Add object assign plugin to properly polyfill `object.assign`. 
  - In #159 we replace an object.assign function, but this PR reverts that commit now that we are properly handling that feature.
* Clean up
   - Adding react version to eslint config to satisfy a warning from the react eslint plugin
   - Updated package.json to point at dr-ui repo rather than old frontend-documentation repo
* Adds eslint-plugin-es to fail if we use any features that we don't have babel set up to compile
   - And turns off any rules that we know to compile as expected. We can adjust these as needed or as we add more babel plugins.

---

To confirm these new Babel changes are working on IE11, I did the following:

1. Ran `npm run build`
2. Locally installed the dr-ui build in the vector-tiles repo.
3. Ran `npm start` and tested in IE11 Browserstack, especially a page using the note component.

---

No hurry in reviewing this PR.


katydecorah:
@colleenmcginnis I updated this PR with master (which now has the erroring plugin from #165 ) and confirmed:

* [x] `npm run build` is successful with /ios
* [x] site loads locally in IE 11
* [x] site loads locally in an older browser (Firefox 44)

PR #165, opened 2019-08-07
[search] polyfill search component to catch errors

katydecorah:
While reviewing Sentry errors, I noticed that the search component is throwing errors in older browser versions like Firefox 44. 

* Adding [babel's poyfill](https://babeljs.io/docs/en/babel-polyfill) to the search component fixes this issue with older browsers versions (just not IE11 because it's out of scope for the search library).
* Ref: https://sentry.io/organizations/mapboxcom/issues/1131166287/?referrer=slack

I built this PR locally and tested on /geographic-analytics:
* [x] IE 11 works 
image

* [x] Firefox 44 works (The search component works , too. I'm not concerned about that weird icon positioning since it doesn't appear in the latest versions of Firefox.)
image

You can test this PR by:
1. npm install
1. npm run build
1. Opening another repo to test locally, like geographic-analytics, and replacing the dr-ui version number in package.json with the path to your local dr-ui repo: `"@mapbox/dr-ui": "file:/Users/katydecorah/Documents/GitHub/dr-ui/pkg",`
1. run npm install (you may need to delete node_modules first) on the tester repo and then npm start
1. open browserstack with Firefox 44 to confirm the site loads locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Master
 * Add a new `Note` style with a green background that will be used to show off new or updated products or features. [#162](https://github.com/mapbox/dr-ui/pull/162)
 * Fix `window.scroll()` on `BackToTopButton`. [#167](https://github.com/mapbox/dr-ui/pull/167)
+* Enable babel polyfill on `Search` component. [#165](https://github.com/mapbox/dr-ui/pull/165)
 
 ## 0.19.3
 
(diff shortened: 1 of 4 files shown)


katydecorah:
@colleenmcginnis looks like adding `babel-plugin-transform-react-remove-prop-types` here as a dependency fixed that error. I'm not exactly sure _why_ yet. My first guess is that it's a peer dependency for one of the two packages also installed in this PR.

PR #166, opened 2019-08-08
Create GL Support wrapper

katydecorah:
This component is a wrapper where we can pass HTML/Component to it as a prop. The wrapper will decide if the browser supports webgl. If it does then it'll show whatever code you passed to it, if it doesn't then it'll display a note with a warning.

Refs https://github.com/mapbox/documentation/issues/236

WebGL supported:
image

WebGL unsupported:
image

How to test:
1. npm install
2. npm start
3. Open the component in a browser you know to have GL support (your regular browser) you should see a typing cat.
4. Open the component in a browser that doesn't support GL or disable GL support. Here's how you can disable GL in Firefox:
    - Open the browser settings page. To do this, type in the address bar: about:config.
    - In the search box, type in: webgl.
    - Find line “webgl.disabled”. It has the “false” value. It is necessary to change it to “true”. To do this simply click twice on this line.

OR!

1. Visit `/help/how-mapbox-works/` on staging
    - I copied this component into /help on this branch https://github.com/mapbox/help/tree/gl-catch and pushed it to staging. 
    - There are [three maps on this page only](https://github.com/mapbox/help/commit/d7f03f162874f87a8c5819b0a80badb28b5c75e0#diff-b2646caced569cf312ab27163dd8b52a) with the wrapper. 
2. Test that page in a browser with GL and without GL support.

 🐌 No rush on review 🐌

### DIFF
--- a/src/components/demo-iframe/demo-iframe.js
+++ b/src/components/demo-iframe/demo-iframe.js
@@ -4,6 +4,10 @@ import ChevronousText from '@mapbox/mr-ui/chevronous-text';
 import GLWrapper from '../gl-wrapper/gl-wrapper';
 
 export default class DemoIframe extends React.Component {
+  static defaultProps = {
+    gl: true
+  };
+
   constructor(props) {
     super(props);
     this.state = {
@@ -36,18 +40,21 @@ export default class DemoIframe extends React.Component {
       ? props.src.replace('MapboxAccessToken', this.state.mapboxAccessToken)
       : props.src;
 
-    return (
-      <GLWrapper>
+    const contents = (
+      <div>
         <iframe src={src} width="100%" height="400px" />
         <a href={src} className="link">
           <ChevronousText text="View fullscreen demo" />
         </a>
-      </GLWrapper>
+      </div>
     );
+
+    return props.gl ? <GLWrapper>{contents}</GLWrapper> : contents;
   }
 }
 
 DemoIframe.propTypes = {
-  src: PropTypes.string.isRequired,
-  MapboxAccessToken: PropTypes.string
+  src: PropTypes.string.isRequired, // absolute URL src for iframe
+  gl: PropTypes.bool, // set to false if the iframe does not use GL
+  MapboxAccessToken: PropTypes.string // optional token value
 };
--- a/src/components/demo-iframe/demo-iframe.js
+++ b/src/components/demo-iframe/demo-iframe.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ChevronousText from '@mapbox/mr-ui/chevronous-text';
+import GLWrapper from '../gl-wrapper/gl-wrapper';
+
+export default class DemoIframe extends React.Component {
+  constructor(props) {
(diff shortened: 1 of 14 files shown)


katydecorah:
@colleenmcginnis Unfortunately, adding `DemoIframe` to d-rui would only fix part of the problem since we load maps in different ways across our sites. For example:
* on this Help page we use the `DemoIframe` a couple times, but we also load a map from a component: https://github.com/mapbox/help/commit/d7f03f162874f87a8c5819b0a80badb28b5c75e0#diff-b2646caced569cf312ab27163dd8b52a
* in /android-docs, we also load a map from its own component: https://github.com/mapbox/android-docs/blob/fc95f6e20f1c9aa3e46b1c1fa229ac4978fd8195/src/components/bearing-tolerance-map.js

### A slight alternative

We could pass children instead of props. I think this implementation looks a bit cleaner:

```
<GLWrapper><DemoIframe src="/help/demos/gl-store-locator/step-five.html" /></GLWrapper>
```

colleenmcginnis:
I guess my main concern is that if this wrapper needs to be added at the Markdown file level for every instance, I'm not sure it will always happen. Maybe we should have a dr-ui-based solution for both use cases: iframes and components.

The iframe component would probably be pretty straightforward (very similar to what we're already using in /help). For Mapbox GL JS maps that are being loaded directly via a component, maybe we should create a basic map component that adds the support wrapper by default. Maybe that's a follow-up task, though? Let me know what you're thinking!

katydecorah:
> this wrapper needs to be added at the Markdown file level for every instance, I'm not sure it will always happen

I see what you're saying now and agree!

> maybe we should create a basic map component that adds the support wrapper by default

I started scoping this out and I think a map component can get pretty complex and would require us to rework all existing GL JS components to use the new component. I think for now, to fix the original issue, adding the GLWrapper around each component is the best option.

Suggested actions:

1. Keep the `GLWrapper` component in this PR, we'll use this for map instantiated in a component.
1. Add the `DemoIframe` component from Help and use the `GLWrapper` within the component.

src/components/demo-iframe/demo-iframe.js
```diff
@@ -0,0 +1,53 @@
...
+        <iframe src={src} width="100%" height="400px" />
+        <a href={src} className="link">
+          <ChevronousText text="View fullscreen demo" />
+        </a>
+      </GLWrapper>
+    );
+  }
+}
+
+DemoIframe.propTypes = {
```
colleenmcginnis: I think we should add another prop here like `requiresGl`, which defaults to `true` and adds the GL wrapper, but can be turned off for the few mapbox.js demos we have and the odd demo like [Create a timezone finder with Mapbox Tilequery API](https://docs.mapbox.com/help/tutorials/create-a-timezone-finder-with-mapbox-tilequery-api/) that doesn't display a GL map. What do you think?

katydecorah:
> I think we should add another prop here like requiresGl, which defaults to true and adds the GL wrapper

💯 thanks so much for catching that use case. I just added the new prop. 

> image
> test cases in a browser with GL disabled

PR #169, opened 2019-08-12
Add a Batfish demo site for testing components 

### LINKED ISSUE #168: Improve testing by adding a batfish testing site in this repo
It would be great to have a directory with a sample batfish build in this repo. 

This would allow us to:

1. Test dr-ui components without having to open another project and install dr-ui locally.
    - What does this solve? Helps reduce time in testing new components locally by having everything in one repo.
2. Add `npm run build` to travis, so that the tests will fail if the batfish build fails.
    - What does this solve? We'll know if dr-ui will fail on `npm run build` sooner

We'll make sure that none of these assets/dependencies for the sample batfish build end up in the packaged version of dr-ui.

katydecorah:
This PR adds a `demo/` directory with a Batfish demo site. 

* We'll use this to test components in context. This will require add the component manually to the demo site. This isn't an ideal workflow right now, but we can revisit.
* The demo app will get built by Travis to assert that the latest dr-ui build doesn't have errors on `npm run build`.

Fixes #168


katydecorah:
@colleenmcginnis It may be a good time to merge this in as well!

* It's all backend and will not affect the package. 
* It would help us catch errors like https://github.com/mapbox/dr-ui/pull/186 (re:  https://travis-ci.com/mapbox/dr-ui/builds/127821556#L342).
* To test components, we won't have to install dr-ui locally in another repo/site.

katydecorah:
I have an idea 💡 ! Going to make the Catalog site run Batfish in #137!

PR #172, opened 2019-08-14
Add syntax highlighting functions

colleenmcginnis:
After reviewing https://github.com/mapbox/android-docs/pull/1037 I realized there were several other docs sites using this pattern. This PR pulls them all together in a central place and adds the same check as in https://github.com/mapbox/android-docs/pull/1037 to each to avoid errors.

@katydecorah I wasn't sure how to test these. I wanted to have something visual so I could confirm that the highlighting looked right. I made a new folder in `src/components/` that only had a `__tests__` folder (no js file with a component) so I could visually test. I feel like this is `wrong` but I think that the build script ignores the `__tests__` folder so maybe it's ok? Let me know what you think! 

screencapture-192-168-137-202-9966-SyntaxHighlighting-2019-08-14-10_15_20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Master
+* Add syntax highlighting helper functions for HTML, Java, JSON, JSX, Objective-C, Swift, and XML. [#172](https://github.com/mapbox/dr-ui/pull/172)
 * Add a new `Note` style with a green background that will be used to show off new or updated products or features. [#162](https://github.com/mapbox/dr-ui/pull/162)
 * Fix `window.scroll()` on `BackToTopButton`. [#167](https://github.com/mapbox/dr-ui/pull/167)
 * Enable babel polyfill on `Search` component. [#165](https://github.com/mapbox/dr-ui/pull/165)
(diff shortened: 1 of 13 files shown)


katydecorah:
> I wasn't sure how to test these.

@colleenmcginnis I think your approach make sense, especially since the tests-cases-app is for us to test what we need. This shouldn't affect the build process.

I'm noticing that there is a lot of repeated code in the `highlight-*` files. What if we consolidated into a single `highlighter.js` file and pass the language to the function? Something like this:

```js
// function
function highlighter(rawCode, language) {
  return rawCode && Prism.languages[language]
    ? Prism.highlight(rawCode, Prism.languages[language])
    : '';
}

// function in action
highlighter(htmlCodeSnippet, 'html')
```

colleenmcginnis:
Just to clarify first -- are you saying only have one function we package up with dr-ui? The reason they were all split up initially what because some require specific Prism `components` (for example, [Java](https://github.com/mapbox/dr-ui/pull/172/files#diff-28e789fd91498bd70a9cdd26520e6a79R2)). Can we account for this using this approach, too?

katydecorah:
@colleenmcginnis hmm yeah maybe one function is a little ambitious, I was hoping to dynamically import the language selected but that seems like a headache.

PR #176, opened 2019-08-29
Create a video component

colleenmcginnis:
Ref https://github.com/mapbox/documentation/issues/254

Creates a `Video` component that will autoplay on loop.

### DIFF
--- a/src/components/video/video.js
+++ b/src/components/video/video.js
@@ -7,8 +7,10 @@ export default class Video extends React.Component {
       autoPlay: true,
       loop: true
     };
-    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
-    const reducedMotion = window !== 'undefined' ? mediaQuery.matches : false;
+    const reducedMotion =
+      window !== 'undefined'
+        ? window.matchMedia('(prefers-reduced-motion: reduce)').matches
+        : false;
     if (reducedMotion) {
       videoProps = {
         autoPlay: undefined,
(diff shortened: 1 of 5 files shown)


colleenmcginnis:
Thanks @katydecorah! I made some updates. When I ran `npm test`, I got an error message "TypeError: window.matchMedia is not a function", which [appears to be a thing](https://stackoverflow.com/questions/39830580/jest-test-fails-typeerror-window-matchmedia-is-not-a-function). Any suggestions? I'm really unfamiliar with these tests. 😬

katydecorah:
@colleenmcginnis 

### For the tests:

Check out: https://stackoverflow.com/a/53449595 We can add this inside each test case to mock the setting. It would be good to have a test that has reduce motion on and off to make sure the attributes update correctly.

We'll also likely want to update package.json to add `"clearMocks": true` to the `jest` config - to clear mocks after each test case.


### For the attributes:

We can store the video's attributes that will change as an object:

```
let videoProps = {
    autoPlay: true,
   loop: true
}
```

And then if the state changes:

```
if (this.state.reduceMotion) {
      videoProps.autoPlay = undefined;
      videoProps.loops = undefined;
      videoProps.controls = true;
    }
```

We can add that object to the element like this: `<video {...videoProps} muted width="100%"... />

src/components/video/video.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class Video extends React.Component {
+  render() {
+    let videoProps = {
+      autoPlay: true,
+      loop: true
+    };
+    const mediaQuery = window.matchMedia('(prefers-reduced-motion: reduce)');
```
katydecorah: You'll want to do a slight swap and make this line evaluate if `window` exists, rather than the next line.

Issue #147, opened 2019-06-24
Add mobile and browser frames


katydecorah:
We should add a library of code-generated mobile and browser frames that screenshots/demos can fit in to add context to the media.

PR #177, opened 2019-08-29
Add Browser component

katydecorah:
image

We can wrap it around any type of media we want to appear in a browser:

```html
<Browser>
  <img src="cool-cats-skateboarding.jpg alt="" />
</Browser>
```

Ref #147 

🐌 no rush


src/components/browser/__tests__/browser-test-cases.js
```diff
@@ -0,0 +1,41 @@
...
+
+const testCases = {};
+
+testCases.basic = {
+  component: Browser,
+  description: '',
+  props: {
+    children: (
+      <img
+        src="https://docs.mapbox.com/studio-manual/assets/ref-styles-manage-styles-1200-599fd585f5f2d8e0729d5fe37f9722c0.webp"
```
colleenmcginnis: This image is not loading for me in Safari.

image

katydecorah: Ohh is that the example image is broken? I'm using a webp image, I should use a png instead!

katydecorah: image

Fixed! I added an `assets` folder in `src/tests-cases-app/` so we can store images/videos/etc that will be bundled only for the test cases site.

colleenmcginnis: oooo yes!

PR #179, opened 2019-09-03
[WIP] iOS and Android toggleable code snippets

colleenmcginnis:
## Background

We have a lot of code throughout docs repos for creating toggleable code snippets especially for Android (Java/Kotlin) and iOS (Swift/Objective-C). I don't think we can use a purely dr-ui component if we want to hook these toggles up to [context](https://reactjs.org/docs/context.html), but here I've modified the `ToggleableCodeBlock` component to be used with a `ContextlessAndroidActivityToggle` and `ContextlessIosViewControllerToggle` component. 

## What's included

- **[new] `CodeSnippetTitle`** so there is a consistent format for adding the filename for a code block (for example, distinguishing between the `Layout` and `Activity` code snippets for Android projects). This also includes an optional `link` prop that, when included, adds a "View on GiHub" link that leads to the relevant GitHub file for additional context.
- **[modified] `CodeToggle`** for toggling between programming languages.
- **[modified] `ToggleableCodeBlock`** programming-language-agnostic component that includes logic for whether or not to include a `CodeSnippetTitle` (only if a `filename` prop is provided) and `CodeToggle` (only if more than one language is provided via the `options` prop).
- **[new] `ContextlessAndroidActivityToggle`** that styles the `ToggleableCodeSnippet` specifically for toggling between Activity code written in Java and Kotlin. This reduces the number of decisions that needs to be made at the Markdown file level.
- **[new] `ContextlessIosViewControllerToggle`** that styles the `ToggleableCodeSnippet` specifically for toggling between ViewController code written in Swift and Objective-C. This reduces the number of decisions that needs to be made at the Markdown file level.
- **[new] `NumberedCodeSnippet`** for displaying line numbers and styling designated `copyRanges` inside snippets. This is currently used in mobile /help tutorials. This is used instead of a generic mr-ui `CodeSnippet` when `copyRanges` are specified.
- **[modified] `prism.css`** to add styling for numbered lines in the `NumberedCodeSnippet` component.
- **[new] `AndroidLayoutCodeBlock`** to make a consistently styled code block that includes auto XML highlighting and an option for adding a `CodeSnippetTitle` (including a link to "View on GitHub").
- **[new] `highlight-kotlin`** to add syntax highlighting for Kotlin code snippets.

## Changes at the repo level

- [ ] The two `Contextless*` components are designed to consume context in each individual repo. We will need to:
    - [ ] Update the context format in each repo.
    - [ ] Develop tests that ensure toggle-language-related context items in individual repos are formatted correctly. 
- [ ] Create context-consuming components in each repo.
- [ ] Modify all instances of code blocks in Markdown files.

## Questions

- How can we import prism.css into the `ToggleableCodeSnippet` component? I get an error when using `raw-loader` (syntax error), but it works as expected when the same content is hardcoded in the component as a string.
- Can we simplify `NumberedCodeSnippet`?

## Next steps

👋  @katydecorah would you be up for looking at the proposed approach? I'll work on push a branch to `ios-sdk`, `android-docs`, and `help` to test in context, then request a more in-depth review. 🙂

### DIFF
--- a/src/components/android-layout-code-block/android-layout-code-block.js
+++ b/src/components/android-layout-code-block/android-layout-code-block.js
@@ -1,3 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CodeSnippet from '@mapbox/mr-ui/code-snippet';
+import CodeSnippetTitle from '../code-snippet-title/code-snippet-title';
+import { highlightXml } from '../../helpers/highlight-xml';
+
+let highlightTheme = `
 code[class*='language-'],
 pre[class*='language-'] {
   color: #273d56;
@@ -113,8 +120,43 @@ pre[class*='language-'] {
 .token.entity {
   cursor: help;
 }
+`;
 
-/* Line numbers */
-[data-content]::before {
-  content: attr(data-content);
+export default class AndroidLayoutCodeBlock extends React.Component {
+  renderTitle = () => {
+    const titleProps = {
+      filename: this.props.filename,
+      link: this.props.link ? this.props.link : undefined
+    };
+    return <CodeSnippetTitle {...titleProps} />;
+  };
+
+  render() {
+    return (
+      <div className="unprose my24">
+        {this.props.filename && this.renderTitle()}
+        {this.props.code && (
+          <div className="round" style={{ backgroundColor: '#f4f7fb' }}>
+            <CodeSnippet
+              code={this.props.code}
+              highlightedCode={highlightXml(this.props.code)}
+              highlightThemeCss={highlightTheme}
+              onCopy={() => {}}
+            />
+          </div>
+        )}
+      </div>
+    );
+  }
 }
+
+AndroidLayoutCodeBlock.propTypes = {
+  /* Optional `filename` to be displayed as a kind of title for the code snippet. */
+  filename: PropTypes.string,
+  /* Optional `link` to a GitHub file. */
+  link: PropTypes.string,
+  /* Raw (unhighlighted) code. When the user clicks a copy
+  button, this is what they'll get. If no highlightedCode
+  is provided, code is displayed. */
+  code: PropTypes.string.isRequired
+};
--- a/src/components/contextless-android-activity-toggle/contextless-android-activity-toggle.js
+++ b/src/components/contextless-android-activity-toggle/contextless-android-activity-toggle.js
@@ -1,0 +1,116 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ToggleableCodeBlock from '../toggleable-code-block/toggleable-code-block';
+import { highlightJava } from '../../helpers/highlight-java';
+import { highlightKotlin } from '../../helpers/highlight-kotlin';
+
+export default class ContextlessAndroidActivityToggle extends React.Component {
+  checkPreference = language => {
+    return this.props.context.preferredLanguage.android === language;
+  };
+
+  render() {
+    const {
+      context,
+      id,
+      java,
+      kotlin,
+      copyRanges,
+      filename,
+      link,
+      limitHeight
+    } = this.props;
+
+    let selectedCode = '';
+    if (context) {
+      selectedCode = this.checkPreference('kotlin') ? kotlin : java;
+    }
+
+    const snippetProps = {
+      copyRanges: copyRanges || undefined,
+      filename: filename || undefined,
+      link: link || undefined,
+      options:
+        kotlin && java
+          ? [
+              {
+                label: 'Java',
+                language: 'java',
+                preferredLanguage: this.checkPreference('java')
+              },
+              {
+                label: 'Kotlin',
+                language: 'kotlin',
+                preferredLanguage: this.checkPreference('kotlin')
+              }
+            ]
+          : undefined
+    };
+
+    return (
+      <div className="my24">
+        <ToggleableCodeBlock
+          {...snippetProps}
+          id={id}
+          code={selectedCode}
+          highlightedCode={
+            this.checkPreference('kotlin')
+              ? highlightKotlin(selectedCode)
+              : highlightJava(selectedCode)
+          }
+          changeLanguage={context.changeLanguage['android']}
+          limitHeight={limitHeight}
+          selectedLanguage={context.preferredLanguage.android}
+        />
+      </div>
+    );
+  }
+}
+
+ContextlessAndroidActivityToggle.propTypes = {
+  /* Intended to be hooked up to `context` in the format
+  used throughout Mapbox docs sites. */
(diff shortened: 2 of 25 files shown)


src/components/toggleable-code-block/toggleable-code-block.js
```diff
@@ -1,52 +1,227 @@
...
-        'javascript'
-      ]),
-      rawCode: PropTypes.string.isRequired,
-      highlightedCode: PropTypes.string.isRequired,
-      preferredLanguage: PropTypes.bool.isRequired
-    })
-  ).isRequired
+  /* A unique `id` is required for the language toggle. */
+  id: PropTypes.string.isRequired,
+  /*  */
```
katydecorah: Is there meant to be a comment here?

src/components/toggleable-code-block/toggleable-code-block.js
```diff
@@ -1,52 +1,227 @@
...
+  selectedLanguage: PropTypes.string.isRequired,
+  /* Raw (unhighlighted) code. When the user clicks a copy
+  button, this is what they'll get. If no highlightedCode
+  is provided, code is displayed. */
+  code: PropTypes.string.isRequired,
+  /* The HTML output of running code through a syntax highlighter.
+  This should use a /helper/higlight-* function. */
+  highlightedCode: PropTypes.string.isRequired,
```
katydecorah: The previous comment on line 205 makes it sound like highlightedCode is optional - should it be required?

src/components/contextless-android-activity-toggle/contextless-android-activity-toggle.js
```diff
@@ -0,0 +1,116 @@
...
+                label: 'Kotlin',
+                language: 'kotlin',
+                preferredLanguage: this.checkPreference('kotlin')
+              }
+            ]
+          : undefined
+    };
+
+    return (
+      <div className="my24">
```
katydecorah: You'll want to add the prose class here so that the code block can get styles from Mapbox Assembly:

```suggestion
      <div className="my24 prose">
```

Right now the test cases app shows a transparent background with minimal styling on the code blocks:

image

src/components/contextless-ios-view-controller-toggle/contextless-ios-view-controller-toggle.js
```diff
@@ -0,0 +1,112 @@
...
+                language: 'objectiveC',
+                preferredLanguage:
+                  this.props.context.preferredLanguage.ios === 'objectiveC'
+              }
+            ]
+          : undefined
+    };
+
+    return (
+      <div className="my24">
```
katydecorah: Like Android, you'll also want the prose class here, too.

```suggestion
      <div className="my24 prose">
```

src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -0,0 +1,323 @@
...
+        const indentingSpacesMatch = line.match(/^[ ]*/);
+        const indentingSpaces = indentingSpacesMatch
+          ? indentingSpacesMatch[0]
+          : '';
+        const indentingSpacesCount = indentingSpaces.length;
+        const paddingLeft = indentingSpacesCount * props.characterWidth + 12;
+        const displayLine = line.replace(/^[ ]*/, '');
+
+        const lineNumberClasses = classnames(
+          'absolute fl color-gray-light align-r pr6',
```
katydecorah: I think we could use color-gray and still achieve the same effect while adding a little more contrast:

```suggestion
          'absolute fl color-gray align-r pr6',
```

src/components/android-layout-code-block/android-layout-code-block.js
```diff
@@ -0,0 +1,162 @@
...
+    };
+    return <CodeSnippetTitle {...titleProps} />;
+  };
+
+  render() {
+    return (
+      <div className="unprose my24">
+        {this.props.filename && this.renderTitle()}
+        {this.props.code && (
+          <div className="round" style={{ backgroundColor: '#f4f7fb' }}>
```
katydecorah: We should add prose here to let assembly style this code block:

```suggestion
          <div className="prose">
```

It'll also prevent the code from overflowing:

image

src/components/contextless-android-activity-toggle/contextless-android-activity-toggle.js
colleenmcginnis: Yes! This was bothering me so much -- I didn't realize `prose` was the solution!

src/components/toggleable-code-block/toggleable-code-block.js
colleenmcginnis: Yea it's not required for the mr-ui component, but I'd like us to always use it in the toggleable code block component. What do you think? I can update the comment above.

katydecorah: I think making it required makes sense!

colleenmcginnis:
>Can we simplify NumberedCodeSnippet?

@katydecorah I first asked this when I thought you were going to work on a simplified version of `CodeSnippet` for /mapbox-gl-js-docs. Now that you're not, it may not be relevant. 

>This component appears to use a lot of the code from mr-ui/code-snippet. There's a lot in there so it's not immediately obvious how these two components are different. I would first look to see how we could use mr-ui/code-snippet if possible.

This is the component that's used in /help for mobile tutorials. I created it because it's not possible to use the mr-ui component to create the same effects. The features that make this component different than mr-ui's `CodeSnippet` include:
- numbered lines
- copyRange highlighting that works well with code snippets with light-colored backgrounds
- scrolling to the first copyRange (which is especially important when there is a maxHeight and the first copyRange is out of view)

We could consider modifying mr-ui's `CodeSnippet` component to add these options, putting out a new release, and updating the version of mr-ui we're using here. Let me know if you think that's the best way to go.

katydecorah:
@colleenmcginnis thanks for the background on the NumberedCodeSnippet, I think it makes sense for us to maintain our own version in dr-ui since it seems the use case is unique to us. So I'm all for continuing development of that component in this PR!

katydecorah:
@colleenmcginnis I have a design suggestion that you're welcome to take or leave! I think by moving the code toggle into the code title, it can help make everything feel more cohesive as a unit:

State | Image
---|---
Before | image 
After | image

To achieve this design we'd need to:
* Adjust code-snippet-title so that the file name and link exist in the same flex element, then using margin-left on the link to give a little extra space
* Add a `toggle` prop to code-snippet
* Pass `renderToggle()` as the value of `toggle` prop to `CodeSnippetTitle` in toggleable-code-block

Let me know what you think! I'd also be happy to help implement if you'd like, but please let me know what you're most comfortable with!

PR #186, opened 2019-09-15
Fix Video build

colleenmcginnis:
When rolling out v0.21.1 to all repos, I came across a build error related to the Video component (catching on [this use of `window`](https://github.com/mapbox/dr-ui/blob/master/src/components/video/video.js#L10-L13)). I did two things in this PR to address it:

- Use /mr-ui's `getWindow` (for consistency).
- Check for the `window` within `componentDidMount` (this seems to be the important part).
  - I chose to store the videoProps (which depend on the results the part that uses `getWindow`) in `state` because that seemed to be a pattern used in mr-ui components that use `getWindow`. Up for other suggestions, though! 

I tested this by building and installing this locally in /ios-sdk and /android-docs. I was able to build both sites successfully and the resulting static site worked as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## master
+
+* Fix build error in `Video` component. [#186](https://github.com/mapbox/dr-ui/pull/186)
+
 ## 0.21.1
 
 * Add `environment` and `location` to the request in `Feedback`. [#184](https://github.com/mapbox/dr-ui/pull/184)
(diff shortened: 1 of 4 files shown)


katydecorah:
🤦‍♂ I didn't double check my work in the last PR and the code to evaluate the window is wrong. We can keep the original code, but use:

```js
const reducedMotion =
      typeof window !== 'undefined'
        ? window.matchMedia('(prefers-reduced-motion: reduce)').matches
        : false;
```

I know I had implemented the use of `@mapbox/mr-ui/utils/get-window` in the feedback module, but I'm rethinking that since if the window is not defined it will throw an error. This isn't necessarily something we want as if the window is undefined, we'd rather it fail silently. I'll create a PR to address this in the feedback module.

PR #189, opened 2019-09-20
new theme for search button

katydecorah:
This PR can be reviewed when we're ready to make the next release.

This PR updates the theme for the Search button:

* Uses Assembly `btn` classes for consistency
* Makes button a little more prominent, without drawing too much attention
* Uses accessible colors (current contrast ratio is too low)

## Default

Status | Image
--- | ----
Before | image
After | image

## Dark

Status | Image
--- | ----
Before | image
After | image


katydecorah:
@colleenmcginnis if we have time to review, this would be good to add in the next release!

katydecorah:
> I'm noticing the formatting gets a little messy in your example right before the break when the search bar jumps onto its own line:

@colleenmcginnis The test case was using code that we normally don't use in the topbarsticker, I updated it and it scales much cleaner!

PR #192, opened 2019-10-02
Add glossary components

colleenmcginnis:
Adds three components for building pages with an alphabetized list of terms (currently used to build the [Glossary page](https://docs.mapbox.com/help/glossary/)):

**`GlossaryCard`** for a single term.

image

**`GlossarySection`** for defining the layout for each letter including the letter and `GlossaryCard`s for relevant cards.

image

**`GlossaryPage`** for defining the layout for all `GlossarySection`s.

image

### DIFF
--- a/src/components/glossary-page/glossary-page.js
+++ b/src/components/glossary-page/glossary-page.js
@@ -1,0 +1,78 @@
+import React from 'react';
+import GlossarySection from '../glossary-section/glossary-section';
+import PropTypes from 'prop-types';
+
+export default class GlossaryPage extends React.PureComponent {
+  makeSectionComponents(termsArr) {
+    let builtSections = [];
+    let currentLetterEntries = [];
+
+    for (let i = 0; i < termsArr.length; i++) {
+      const thisLetter = termsArr[i].title.charAt(0).toLowerCase();
+      const nextLetter =
+        i === termsArr.length - 1
+          ? termsArr[i].title.charAt(0).toLowerCase()
+          : termsArr[i + 1].title.charAt(0).toLowerCase();
+
+      if (thisLetter < nextLetter) {
+        currentLetterEntries.push(termsArr[i]);
+        const thisLetterSectionComp = (
+          <GlossarySection
+            key={thisLetter}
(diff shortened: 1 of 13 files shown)


src/components/glossary-page/glossary-page.js
```diff
@@ -0,0 +1,78 @@
...
+            />
+          );
+          builtSections.push(thisLetterSectionComp);
+        } else {
+          currentLetterEntries.push(termsArr[i]);
+        }
+      }
+    }
+    return builtSections;
+  }
```
katydecorah: I think we can trim down this function a little bit to make it easier to make future updates. This code may need some work (and we'll want to double-check that it works in IE11), but we could potentially replace the alphabetizedTerms with:

```js
// sorting function that is case insenitive
const sortBy = key => (a, b) =>
  a[key].toLowerCase() > b[key].toLowerCase() ? 1 : b[key].toLowerCase() > a[key].toLowerCase() ? -1 : 0;

// alphabetized the list of terms
const alphabetizedTerms = this.props.entries.reduce((arr, entry) => {
  // get the first letter
  const letter = entry.title[0].toLowerCase();
  // find all the entries that start with "letter" and then sort by title
  const entries = this.props.entries.filter(e => e.title[0].toLowerCase() === letter).sort(sortBy('title'));
  // if the letter doesn't exist in "arr" yet, push the letter with matching entries
  if (!arr.filter(l => l.letter === letter).length) arr.push({letter, entries})
  return arr;
}, []).sort(sortBy('letter'))
```

and then update makeSectionComponents with:

```js
makeSectionComponents(terms) {
    return terms.map(term => {
      return (
        <GlossarySection
          key={term.letter}
          letter={term.letter}
          entries={term.entries}
        />
      );
    });
  }
```

It would also be good to include a lowercase title to make sure it sorts correctly!

colleenmcginnis:
Thanks @katydecorah! This code worked well with my test case (I also checked in IE11). Are there other test cases we should include (entries that start with numbers comes to mind, but there could be others).

PR #195, opened 2019-11-06
Add Android and iOS frames

katydecorah:
I think I was shy about this PR because the implementation is rather funky: it uses Assembly and inline styles to create the frames so we don't have to depend on an external stylesheet. I'm not sure if this makes the components more complicated.

Platform | Portrait | Landscape
---|---|---
iOS | image| image
Android | image |  image

Getting this PR up for visibility, but still need to do some testing, specifically:

* [x] Check to make sure frames fit around android and ios videos and images
* [x] Determine what should happen to the frames on smaller screen: do they scale? 
* [x] How does the notch looks on ios videos and images? Should we blank out the space around the notch or keep it transparent?
* [x] Does the gray color of the frames hold up in practice?

cc @colleenmcginnis 

Fixes https://github.com/mapbox/dr-ui/issues/147


katydecorah:
Gosh, I wonder if straight CSS would just be better. This seems sooo complicated.

katydecorah:
bg-gray is clashing a bit, maybe we should try bg-gray-dark:

image

also need to adjust the height to prevent the gap at the bottom

katydecorah:
Instead of passing an image or video path, I wonder if it should accept a node so that we can pass AppropriateImage or Video to the frame.

katydecorah:
Getting there!

image

The video is getting a little clipped here, would be great if it snapped to the asset's dimensions instead.

katydecorah:
Replace the height for a min-height and now the frame will snap around the asset:

image

image

Also hiding the buttons for the landscape phones on mobile because they don't scale as well.

katydecorah:
I sent a lot of time debugging an IE11 error which wouldn't have happened if this was using straight CSS. For the sake of future us, I'm going to refactor this to use CSS.

katydecorah:
Notch on iOS is looking good:

image

katydecorah:
This is ready for review! There is a branch called `dr-ui-phone` in android and ios repos where I updated the `VideoWithDeviceFrame` component to use this new component. You'll need to install dr-ui locally to see the changes.

src/components/phone/phone.js
```diff
@@ -0,0 +1,221 @@
...
+                className={config.notchCurveRightClasses}
+                style={config.notchCurveRightStyles}
+              />
+            )}
+          </div>
+        </div>
+        {/* screen */}
+        <div style={config.screenStyles} className={config.screenClasses}>
+          <div className="loading loading--dark absolute z0" />
+          <div className="relative">{props.children}</div>
```
colleenmcginnis: @katydecorah in addition to swapping out the device frames for videos in your test branches, I also tested using this around images (because it would be great for all images and videos to be consistent!).

However, it seems like the logic in `AppropriateImage` might think the container width is 0px, so it's always rendering the smallest image (even when the screen is wide enough that it should be rendering a larger version). After the wrong (too small) image is rendered, if i resize the browser it recalculates and swaps in the right (bigger) version of the image.

I was able to get around this by adding `w-full` to this `div`. In my testing, it didn't seem to have any negative impacts, but I'm curious what you think about this solution!

katydecorah: @colleenmcginnis would you mind pushing an example of AppropriateImages to one of the test branches? I haven't been able to recreate this yet 🙏

PR #196, opened 2019-11-27
Use @mapbox/eslint-config-docs

katydecorah:
* [x] `npm install @mapbox/eslint-config-docs --save-dev && npx install-peerdeps --dev @mapbox/eslint-config-docs`
* [x]  remove (now) redundant fields in conf/eslint-config-browser.json
* [x]  add `"@mapbox/eslint-config-docs"` to extends while removing an extended modules already present in eslint-config-docs
* [x] fix eslint errors

Ref https://github.com/mapbox/documentation/issues/272


src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -50,7 +50,7 @@ export default class NumberedCodeSnippet extends React.PureComponent {
     injectedThemes.push(theme);
     const doc = getWindow().document;
     this.styleTag = doc.createElement('style');
-    this.styleTag.innerHTML = theme;
+    this.styleTag.innerHTML = `${theme}`;
```
colleenmcginnis: Curious about the reason for this rule. 🤓

src/helpers/create-redirect.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 export function createRedirect(target) {
   return class Redirect extends React.Component {
     componentDidMount() {
-      window.location.href = target;
+      window.location.href = escape(target);
```
colleenmcginnis: Also this!

katydecorah: This rule is from https://github.com/mapbox/eslint-config-mapbox/blob/master/xss.js - this adds a layer of protection by escaping the url. This prevent hacks! (Which are unlikely in this environment since we define the `target` but it's good practice.

src/components/numbered-code-snippet/numbered-code-snippet.js
katydecorah: This rule is also from https://github.com/mapbox/eslint-config-mapbox/blob/master/xss.js 

https://github.com/Rantanen/eslint-plugin-xss/blob/master/docs/rules/no-mixed-html.md

If we were dealing with user input, we'd want to escape this value, but was able to wrap in a template literal instead.

PR #200, opened 2019-12-11
Allow for a configurable connector on search component

rclark:
This PR adjusts the search component, allowing you to provide a custom `SiteSearchAPIConnector`. This allows the same component to be used to search other Swiftype "engines".

## Todo

- [x] write a test for the condition
- [x] look into whether prop types can assert that the `.connector` is an `instanceOf SiteSearchAPIConnector`


src/components/search/__tests__/search.test.js
```diff
@@ -119,4 +119,20 @@ describe('search', () => {
...
+
+    beforeEach(() => {
+      testCase = testCases.withLayout;
+      wrapper = renderer.create(testCase.element);
+      tree = wrapper.toJSON();
+    });
+
+    test('renders as expected', () => {
+      expect(tree).toMatchSnapshot();
+    });
```
rclark: @katydecorah any suggestions on how I might make some kind of assertion that the connector provided to the test case is actually in use? It shouldn't actually change anything about how the component renders, only how it behaves.

src/components/search/search.js
```diff
@@ -25,7 +25,7 @@ class Search extends React.Component {
     ) : (
       <SearchProvider
         config={{
-          apiConnector: connector,
+          apiConnector: props.connector || connector,
```
rclark: This change is backwards-compatible.

src/components/search/__tests__/search.test.js
```diff
@@ -119,4 +119,20 @@ describe('search', () => {
...
   });
+
+  describe(testCases.withConnector.description, () => {
+    let testCase;
+    let wrapper;
+    let tree;
+
+    beforeEach(() => {
+      testCase = testCases.withLayout;
+      wrapper = renderer.create(testCase.element);
```
katydecorah: We'll want adjust these lines to pull in the `withConnector` test case and also since the `withConnector` test case is formatted as props (not an element like the `withLayout` test case) we'll need to use React's `createElement` to build the `wrapper`.

```
testCase = testCases.withConnector;
wrapper = renderer.create(
  React.createElement(testCase.component, testCase.props)
);
```

After you make this change, you'll need to update the jest snapshots with `jest -u` and you can commit those files. The diff for this test case will look lengthy, but that's only because we're removing the generated snapshot code that got pulled in from the `withLayout` test case.

src/components/search/__tests__/search.test.js
katydecorah: Perhaps we could use https://jestjs.io/docs/en/expect#tobeinstanceofclass?

Something like:

```
expect(testCase.props.connector).toBeInstanceOf(SiteSearchAPIConnector);
```

src/components/search/search.js
katydecorah: One suggestion that would help keep our props organized is to add `connector` to defaultProps (down on line 103) with the value of doc's engine (lines 9-13).

Then on this like we can change to:
```suggestion
          apiConnector: props.connector,
```

defaultProps is where we define the initial values and then if the component receives a specific prop value then it will override with the value passed to that component.

src/components/search/__tests__/search.test.js
rclark: oops!

src/components/search/__tests__/search.test.js
rclark: Added two expectations -- the instance of check and also a check on the connector's properties to make sure its not using the default values.

rclark:
Thanks @katydecorah for the feedback. This all feels a lot better to me now. Final review?

PR #202, opened 2019-12-13
Rework Notes themes

### LINKED ISSUE #175: Rethink how we're theming Notes
In at least two separate repos, we have components for:
* download-note.js
* legacy-note.js
* warning-note.js
* note.js

I'm wondering if we should audit the types of notes and create themes that automatically include the SVG icon. For example:

Before | After
--|--
```<Note theme="warning" imageComponent={<WarningImage color="orange" />}>This is a legacy product...</Note>``` | ```<Note theme="legacy">This is a legacy product...</Note>```

### PR DESCRIPTION
**Pull request by katydecorah**
This PR will rework themes by removing the `imageComponent` and building the image with the theme. This PR also:
* removes new-image, warning-image - since they are already Assembly icons, we can remove this components and create an embedded component to handle the styling
* maps together similar themes

Proposed theme set:
* 💙Default
* 🧡Warning, Legacy (these notes already use the same theme so both themes will return the same style)
* ❤️Error
* 💜Download
* 💚Beta, New (these notes already use the same theme so both themes will return the same style)

To do:
* [x] option for no image
* [x] finish auditing notes used across mapbox.com to make sure we've covered all use cases.

fixes #175

### CODE REVIEW

File: src/components/note/note.js [hunk 1]
```diff
@@ -1,60 +1,98 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+import BookImage from '../book-image/book-image';
```
colleenmcginnis: What do you think about using the [`book` icon from Assembly](https://labs.mapbox.com/assembly/icons) here instead? I do like the lil book image, but I also like consistency!

src/components/note/note.js
```diff
@@ -1,60 +1,98 @@
...
+          {children}
         </div>
       </div>
     );
   }
 }
 
+Note.defaultProps = {
+  theme: 'default',
+  title: 'Note',
```
colleenmcginnis: What do you think about different default titles for the different themes? Is that unnecessarily complicated? For example `Public Beta` for `beta` and `Download` for `download`.

src/components/note/note.js
katydecorah: Added!

image

src/components/note/note.js
katydecorah: Super smart! Done!

image

PR #206, opened 2019-12-16
Collapse grayed lines in code block

colleenmcginnis:
This proposes the ability to collapse or expand the grayed lines not included in `copyRanges`. This functionality would be useful in tutorials where we want to show the full code for context, but want to highlight a few lines that are relevant to the step being described in the text. 

## Default state

Screen Shot 2019-12-16 at 2 33 45 PM

## After clicking "Show all lines"

Screen Shot 2019-12-16 at 2 34 12 PM

@mapbox/docs thoughts?


src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -106,7 +114,7 @@ export default class NumberedCodeSnippet extends React.PureComponent {
       overlayElement.style.opacity = '1';
       copyElement.style.opacity = '1';
     }
-  }, 300);
+  }, 10);
```
colleenmcginnis: @katydecorah when this was set to `300`, there was a very noticeable delay of the blue bar on the left side moving to the relevant lines when expanding/collapsing the grayed out lines. Do you think we need to use `debounce` when expanding/collapsing lines (especially since we won't need to auto scroll to the first active chunk anymore)?

colleenmcginnis: Oh I have an idea. 💡

colleenmcginnis: @katydecorah @danswick I think `debounce` was introduced in the mr-ui code snippet component specifically for recalculating the position of elements when the window is resized. I'm not sure we need to use it when toggling between languages or expanding/collapsing lines since it each of those triggers happen only once. What do you think of an approach like this: https://github.com/mapbox/dr-ui/pull/206/commits/30db2d675a1370ae6ce7a2647c0aae8a5fbe4cef?

katydecorah: @colleenmcginnis your solution is super smooth! Looks good to me!

colleenmcginnis: 🏄‍♀

katydecorah:
Noticing one small issue on smaller screens where the code overflows:

image

colleenmcginnis:
@katydecorah I think there may be some situations in which we don't want to collapse the grayed out lines (thinking about the installation flow) so in https://github.com/mapbox/dr-ui/pull/206/commits/10172856938fce78f0bcccdac5594ba011325d76 I added a prop that would give you the option to specify that lines should not be collapsed. Let me know what you think!

katydecorah:
@colleenmcginnis should the code snippet scroll to the highlighted code? I'm not seeing it do that for this example: image

colleenmcginnis:
You know. It's weird. When I comment out these lines for the `basic` test https://github.com/mapbox/dr-ui/blob/10172856938fce78f0bcccdac5594ba011325d76/src/components/numbered-code-snippet/__tests__/numbered-code-snippet-test-cases.js#L167-L178 so the second test case is the only one on the page, it works as expected. I'm not sure if it's an issue specific to how we're setting up test cases or if there's something else going on that we should be concerned about.

katydecorah:
I'll take a closer look to see what's up! We want to make sure the test cases work as expected because it's hard to track use cases like this.

src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -11,52 +12,94 @@ function getWindow() {
...
     const theme = this.props.highlightThemeCss;
-
-    // Do not load themes that have already been injected.
+    /* Do not load themes that have already been injected. */
     if (injectedThemes.indexOf(theme) !== -1) return;
     injectedThemes.push(theme);
     const doc = getWindow().document;
     this.styleTag = doc.createElement('style');
     this.styleTag.innerHTML = `${theme}`;
     doc.head.appendChild(this.styleTag);
```
katydecorah: @colleenmcginnis I think these lines are the culprit and returning too soon for the second snipped instance to scroll. I recommend moving these lines to is own method:

```
  setTheme = () => {
    const theme = this.props.highlightThemeCss;
    /* Do not load themes that have already been injected. */
    if (injectedThemes.indexOf(theme) !== -1) return;
    injectedThemes.push(theme);
    const doc = getWindow().document;
    this.styleTag = doc.createElement('style');
    this.styleTag.innerHTML = `${theme}`;
    doc.head.appendChild(this.styleTag);
  }
```

And then referencing that method here: `this.setTheme()`

PR #209, opened 2019-12-17
Clean up feedback -> Sentry tags

katydecorah:
This PR:

* conditionally sets `section` and `preferredLanguage` (this will prevent the  tag value of `undefined` if the values are unset)
* initialize the docs-feedback Sentry project with the environment (production | staging) so that we can keep our project tidy. (We'll be able to filter by environments in Sentry)
* allow `feedbackSentryDsn` to accept boolean (false) to disable sending text feedback to Sentry (this is added with Platform team in mind)

🐌


katydecorah:
@colleenmcginnis I'd like to include this in the next release, if you don't mind 👀

src/components/feedback/feedback.js
```diff
@@ -7,6 +7,12 @@ import Icon from '@mapbox/mr-ui/icon';
 import * as Sentry from '@sentry/browser';
 
 const anonymousId = uuidv4(); // creates an anonymousId fallback if user is not logged or we cant get their info
+const environment =
```
colleenmcginnis: I don't understand what the resulting `environment` does in line 48, but maybe it's just that I don't understand how `Sentry.init()` works. 🤔

katydecorah: @colleenmcginnis I can see how it's not clear as it's using shorthand on line 48. We're setting the key `environment` in the Sentry.init object to the value of `environment` as declared here. Which could also look like this (but when the value and key are the same, you can set the variable without having a key/value pair):

```js
Sentry.init({
  dsn: this.props.feedbackSentryDsn,
  environment: environment
});
```

PR #212, opened 2019-12-19
Add "tag" option to second and third level headings

colleenmcginnis:
Adds an optional `tag` prop to `secondLevelHeadings` and `thirdLevelHeadings` in `NavigationAccordion`.

Screen Shot 2019-12-19 at 8 22 57 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Add IE11 compatibility for the `Search` component.
 * Update `BetaFlag` component styling. [#207](https://github.com/mapbox/dr-ui/pull/207)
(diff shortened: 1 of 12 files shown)


colleenmcginnis:
Excellent points! I have a slight preference for `tag` over flag, but could be convinced otherwise. I can merge this PR and #207 first, then work on these two points in a follow up PR. 🎉 

>Then this tag prop would accept a string for the theme of the flag instead of a whole component.

I'm slightly hesitant to disallow nodes altogether in case we're moving fast and need to add a new tag theme to one repo temporarily until the next dr ui release can be done. Maybe there's a way to allow a "custom" option + component for those temporary situations. 🤷‍♀

katydecorah:
Sounds good!

PR #216, opened 2019-12-20
Add a tag component with themes

colleenmcginnis:
⚠️  **Includes breaking changes to the following components:**
- **`BetaFlag`**
- **`OverviewHeader`**
- **`NavigationAccordion`**
- **`ProductMenu`**

Introduces a generic `Tag` component with several available themes. I removed `BetaFlag`. You can use `<Tag theme="beta" />`  instead. Themes align with `Note` themes in #202. 

image

PR #217, opened 2019-12-20
Create themes component to house shared themes for Tag and Note

katydecorah:
Blocked by #216 

This PR adds a Theme "component",  using quotes here since it exports a JSON object, that stores shared theme data for `Tag` and `Note`. This PR also tests the color contrast between the background color and color in Themes to help make sure they are accessible.

To test:
1. npm ci
2. npm start
3. navigate to the /Themes page to see each note and, when available, coordinating tag.


src/components/themes/themes.js
```diff
@@ -0,0 +1,97 @@
...
+      borderColor: '#fd8ac0'
+    }
+  }
+};
+/* duplicate themes */
+themes.legacy = Object.assign({}, themes.warning);
+themes.legacy.label = 'Legacy';
+(themes.legacy.tooltipText =
+  'This feature is no longer in active development.'),
+  (themes.new = Object.assign({}, themes.beta));
```
colleenmcginnis: Now that the default theme is gray (and blue is available), what do you think about using blue for either `new` or `beta`?

katydecorah: wowowowowow blue for beta!!!!!!!

katydecorah: I like the blue! Just want to flag that the color of links is the same color as the text:

image

@colleenmcginnis I think it's fine since links are underscored, but wanted to double check with you!

katydecorah: Ohhh and icon, maybe plus symbol for new?

image

colleenmcginnis: >@colleenmcginnis I think it's fine since links are underscored, but wanted to double check with you!

I think /help pages have custom CSS that removed the underline 😬 but maybe we should change that for consistency anyway!

katydecorah: @colleenmcginnis yeah let's make sure /help underscores links since that's a best practice 🤠

PR #197, opened 2019-12-04
Create `Edit` and `CodeSnippet` components

katydecorah:
This PR adds a new component that you can send code and metadata to and these buttons will bring it into your favorite editor:

image

This PR also adds CodeSnippet that include:
- option to add Edit in buttons
- option add CodeSnippetTitle
- automatically pulls in prism.css

To do:
* [x] more tests, especially to catch errors with prettier attempting to parse bad html
* [x] break code into smaller functions, add tests
* [x] rework props to simplify (html, css, js, resources, frontmatter) we are currently doing a lot of massaging and pieces together of data to get it how we want, instead we should provide helper functions to help extract data so that this component can stay simple

## How to test /mapbox-gl-js-docs:
1. from this branch, `npm run build` (build it locally)
2. in mapbox-gl-js-docs, pull down the `dr-ui-edit-buttons` branch
3. in mapbox-gl-js-docs, install dr-ui locally: `"@mapbox/dr-ui": "file:/Users/katydecorah/Documents/GitHub/dr-ui/pkg",` (update this with your path to your local dr-ui)
4. in mapbox-gl-js-docs, `yarn install` then `yarn start`
5. navigate to an examples page and test any of the edit buttons

## How to test in /help:

1. from this branch, `npm run build` (build it locally)
2. in help, pull down the `dr-ui-code-snippet` branch
3. in help, install dr-ui locally: `"@mapbox/dr-ui": "file:/Users/katydecorah/Documents/GitHub/dr-ui/pkg",` (update this with your path to your local dr-ui)
4. in help, `npm install` then `npm start`
5. navigate to http://localhost:8080/help/tutorials/use-mapbox-gl-js-with-react/  and see the CodeSnippet in action

PR #218, opened 2020-01-03
create highlight/theme-css

katydecorah:
We have copied and pasted prism.css in a few places to avoid issues with importing the css file at src/css/prism.css.

This PR centralizes the code base for prism.css by creating a build script that will:
1. take the contents of the file at `src/css/prism.css`
2. create a new file at `src/components/highlight/theme-css.js`. This new file exports a constant, where the value is the contents of prism.css. Example:

```js
import { highlightThemeCss } from '@mapbox/dr-ui/highlight/theme-css.js';

<CodeSnippet highlightThemeCss={highlightThemeCss} />
```

The script will run at prestart and prebuild to make sure the file is generated in time.


katydecorah:
@colleenmcginnis I updated the original comment to better explain this addition!

I was inspired to add this as I was working on #197 CodeSnippet and didn't want to copy and paste prism.css another time. For our fancy CodeSnippet, I think we should bake the css theme into the component and not make it a prop since I can't think of an instance where we use a different set of styles.

PR #220, opened 2020-01-03
Update dependencies

katydecorah:
Before we make the next release, it would be good to get on the latest version of a couple dependencies.


katydecorah:
gonna sneak in a prismjs update!

katydecorah:
The latest commit updated prismjs to the latest. A few snapshots were updated as it looks like the latest version is wrapping more syntax. Here's an example of on of the changes:

Current:
image


Proposed:
image

PR #234, opened 2020-01-27
create playground theme for RelatedPage

### LINKED ISSUE #231: Add `playground` theme for RelatedPage
- I noticed a few RelatedPage instances in /android where it linked to a playground - so we set it to the default.
- We could use this component in /api to highlight a playground for an API. Right now it's a small callout: 
image

cc @colleenmcginnis

katydecorah:
This PR adds a `playground` theme/contentType for RelatedPage. Example in use:

> image

## QA Checklist

fixes #231


katydecorah:
I went with the default theme only because the older interactive-tool.svg was already using those colors. I believe purple and yellow are left as color options - I do not have an opinion and am happy to update the icon/theme if someone does :)

src/components/related-page/related-page.js
```diff
@@ -7,6 +7,45 @@ import BookImage from '../book-image/book-image';
 import GlossaryImage from '../glossary-image/glossary-image';
 import ExampleImage from '../example-image/example-image';
 
+class PlaygroundImage extends React.Component {
```
colleenmcginnis: Should we put `PlaygroundImage` into its own component and import to be consistent with `GlossaryImage`, `ExampleImage`, etc. above?

katydecorah: @colleenmcginnis my initial thought was that we likely wouldn't use the PlaygroundImage elsewhere, but that could change. I'll break it out into its own component!

katydecorah:
@colleenmcginnis making PlaygroundImage was a good call, I found a lil bug with the svg. This is ready for review again!

PR #255, opened 2020-02-20
Send basic user metadata along with Feedback

katydecorah:
This PR:
- Replaces the Feedback component's `userName` prop with the `user` object (which we will populate using a docs-page-shell function)
- Sends `plan` information to Segment 
- Sends user data (username, plan, email) to Sentry
- Adds line comments

To do:

* [x] Decide if `plan` should be a tag or part of the user object.

## How to test

1. `npm start` and open /Feedback test case app OR test the /api-documentation branch `dr-ui-0.27.0` locally
2. You should check the response in two places:
   - #bots (Segment) 
   - [docs-feedback Sentry project](https://sentry.io/organizations/mapboxcom/issues/?environment=staging&project=1848287&referrer=slack)

On the /api-documentation branch, try sending feedback signed in and signed out. Look for:
- signed in - Your userId and plan are found in the #bots event object. Your userId and email are found in the "User" section of the Sentry issue and your plan is set as a tag.
- signed out - No user information is passed, only an anonymousId.

## QA Checklist

- [x] No errors logged to console.
- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `5.7.0`.
    - Score should be 89, due to #257 #248 and the `#docs-content` id that's baked into the test-cases app
- [x] Component is accessible at mobile-size viewport.
- [x] Component is accessible at tablet-size viewport.
- [x] Component is accessible at laptop-size viewport.
- [x] Component is accessible at big-monitor-size viewport.
- [x] If applicable, create a PR in a site repo, copy the component, and test it. Push to staging and instruct the reviewer to test the component in-page.

Open the test cases app locally on:

- [x] Chrome, no errors logged to console.
- [x] Firefox, no errors logged to console.
- [x] Safari, no errors logged to console.
- [x] Edge, no errors logged to console.
- [x] IE11, no errors logged to console.
- [x] Mobile Safari, no errors logged to console.
- [x] Android Chrome, no errors logged to console.

Issue #248, opened 2020-02-10
NavigationDropdown, SectionedNavigation - Form elements do not have associated labels


katydecorah:
- [ ] The `ControlSelect` in `NavigationDropdown` is missing an associated label.
- [ ] The `ControlInput` in `SectionNavigation` is missing an associated label.

https://web.dev/label/

PR #256, opened 2020-02-21
Add tests to assert Segment event in Feedback component

katydecorah:
This PR:
- adds tests to assert the object shape for events sent to Segment.
- pushes the Segment event to the state for easier testing/tracking.
- adds createId function to generated a unique id for each interactive element.
- reorganizes functions and adds comments to help make all the actions clear.

I think this PR will help us better track additions to metadata (like in #255).

## How to test

1. `npm start`
2. Open /Feedback
3. Try any test case, you can see the Segment event posted in #bots and the Sentry feedback posted in #docs-feedback.

## QA Checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `7.5.0`.
  - Score is 89 due to: 
    - #257 color contrast 
    - [id] attributes on the page are not unique (#docs-content is baked into test cases)
    - #248 form elements missing label
- [x] If applicable, create a PR in a site repo, copy the component, and test it. Push to staging and instruct the reviewer to test the component in-page.

### DIFF
--- a/src/components/feedback/feedback.js
+++ b/src/components/feedback/feedback.js
@@ -5,6 +5,7 @@ import forwardEvent from './forward-event';
 import uuidv4 from 'uuid/v4';
 import Icon from '@mapbox/mr-ui/icon';
 import * as Sentry from '@sentry/browser';
+import slugify from 'slugify';
 
 const anonymousId = uuidv4(); // creates an anonymousId fallback if user is not logged or we cant get their info
 const environment =
@@ -30,8 +31,13 @@ class Feedback extends React.Component {
     this.sendToSegment = this.sendToSegment.bind(this);
   }
 
+  // creates a unique id for an element
   createId = el => {
-    return `dr-ui--feedback-${this.props.section || 'page'}-${el}`;
+    const section = slugify(this.props.section || 'page', {
+      replacement: '-',
+      lower: true
+    });
+    return `dr-ui--feedback-${section}-${el}`;
   };
 
   // pushes the text feedback to the state as the user types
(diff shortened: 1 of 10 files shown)


katydecorah:
@danswick this is ready for review (the dr-ui release isn't due until next Friday, so not urgent today). As I was working on #255 I felt weird about not having tests so this PR adds tests!

src/components/feedback/feedback.js
```diff
@@ -13,85 +14,129 @@ const environment =
...
       helpful: undefined,
       feedback: undefined,
-      feedbackSent: undefined
+      feedbackSent: undefined,
+      event: undefined
     };
     this.handleText = this.handleText.bind(this);
     this.handleYesNo = this.handleYesNo.bind(this);
-    this.submitFeedback = this.submitFeedback.bind(this);
+    this.handleSubmitFeedback = this.handleSubmitFeedback.bind(this);
```
danswick: What do you think about converting all of these to arrow functions? Since we're already using the babel `transform-class-properties` plugin in this repo, so I think we should be covered when it builds!

src/components/feedback/feedback.js
```diff
@@ -13,85 +14,129 @@ const environment =
...
-    // sends event to segment via forward event webhook
-    forwardEvent(event, this.props.webhook, err => {
+  }
+
+  // sends event to Segment
+  sendToSegment() {
+    // sends event to Segment via forward event webhook
+    forwardEvent(this.state.event, this.props.webhook, err => {
       if (err) {
         console.log(err); // eslint-disable-line
```
danswick: Do you think it's worth exposing these errors so we can log them in sentry and/or provide some visual feedback to the user?

src/components/feedback/feedback.js
katydecorah: Yes! Good thinkin!

src/components/feedback/feedback.js
katydecorah: Ohhh yes, we should!

PR #262, opened 2020-03-04
[draft/proposal] create passiveSection option for NavigationAccordion

katydecorah:
When working with content and NavigationAccordion, we use the page's hierarchy to populate the navigation. There are cases, especially in very big documents, that the page's hierarchy doesn't always match with how we want to present the navigation.

This PR proposes adding an optional feature to enable small headings within a section:

- These headings do not link to content and only exist to help group together content.
- These passive headings allow us to surface the most important headings.

In the example below, we would ordinarily need to make "Methods and events" and "Properties" the h2s on the page, but that would place the other headings as h3s, which are hidden until the user scrolls to that section. The h3s are actually the most compelling content, so by adding this passiveSection feature, we can help organize the content without sacrificing with how we're marking up the page:

> image

## QA checklist

- [ ] IE11, no errors logged to console.

### DIFF
--- a/src/components/feedback/__tests__/__snapshots__/feedback.test.js.snap
+++ b/src/components/feedback/__tests__/__snapshots__/feedback.test.js.snap
@@ -171,13 +171,13 @@ exports[`feedback Feedback placement (sends anonymous user information) renders 
                         </div>
                       </a>
                       <div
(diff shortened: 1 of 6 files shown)


katydecorah:
I don't think we need this anymore

PR #268, opened 2020-04-02
Fix bug in NumberedCodeSnippet

colleenmcginnis:
This PR is primarily fixing a bug in `NumberedCodeSnippet` where if the last line of the code snippet being used was included in a `copyRange`, the chunk would be marked as `live: false` when it should be `live: true`.

The only approach I could get to work was to loop through all lines _plus_ one more time to give it the chance to `endCurrentChunk` after the last line, but not push anything to the `currentChunk` the last time it loops through. I have a hunch @katydecorah might be able to come up with a more elegant solution, but this one seems to work!

In addition to this bug fix, the PR also:
- Adds test cases for when the `copyRange` includes the first line and the last line.
- Prevents showing Hide/Show lines if there are no line elements (this was happening at the beginning and end of code snippets).

## How to test

- Run `npm start`.
- Visually check all test cases for `NumberedCodeSnippet`.
    - The "First line" test case should display the first five lines of the sample code snippet and only include `Show hidden lines` at the bottom of the live chunk.
    - The "Last line" test case should display the last five lines of the sample code snippet and only include `Show hidden lines` at the top of the live chunk.

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `#.#.#`. (will be 91% until https://github.com/mapbox/dr-ui/pull/266 merges)
- [x] Create a PR in a site repo, copy the component, and test it. Push to staging and let the reviewer know they can also test the component there. (tested myself in /help 😬)

cc @chezzdev


src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -217,22 +217,81 @@ export default class NumberedCodeSnippet extends React.PureComponent {
...
+         * the allChunks array as a live chunk.
+         */
+        endCurrentChunk({ live: true });
+      } else {
+        /**
+         * If a currentLiveRange does not exist, end the current chunk and add
+         * it to the allChunks array as a non-live chunk.
+         */
+        endCurrentChunk({ live: false });
+      }
```
colleenmcginnis, line 294: @katydecorah while I was writing code comments, I realized there was a cleaner way to do this and changed my approach. Can you take a second look? 

Basically, I reverted all other changes I made and handled it here. Hopefully the code comments thoroughly explain my approach, but please let me know if you have any questions!

src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -217,22 +217,81 @@ export default class NumberedCodeSnippet extends React.PureComponent {
...
       currentChunk.push({
         highlighted: chunk,
         raw: rawCodeLines[i]
       });
     }
+
+    /**
+     * After iterating through all lines, if the currentChunk still contains
+     * contents that means the chunk has not yet been added to the allChunks
+     * array (which is doen with endCurrentChunk).
```
colleenmcginnis, line 274:
```suggestion
     * array (which is done with endCurrentChunk).
```

PR #274, opened 2020-05-01
add `isStuck` prop to `TopBarSticker` component

katydecorah:
I would like to propose adding an `isStuck` prop to `TopBarSticker`. This prop's default is `true and allows the option to set it to `false` for occasions where we do not want a sticky top bar.

I'm keen on making the TopBarSticker static on /playground since there aren't any navigational items in the top bar and it takes up valuable screen space. Ref https://github.com/mapbox/playground/issues/78

🐌 

## How to test

1. Run `npm start`
2. open /TopbarSticker test case
3. The last test case should _not_ stick.

## QA checklist

- [ ] IE11, no errors logged to console.


colleenmcginnis:
This is a great idea. I think we could also use it in /help where we're using custom code to recreate a version of this component that doesn't stick https://github.com/mapbox/help/blob/publisher-production/src/components/site-navigation.js#L9.

katydecorah:
> I almost wonder if we should have a separate TopBar and TopBarSticker component. Do you think there's any benefit to separating out the styling of the top bar from the behavior made possible by react-stickynode? (Especially if we might have more use cases where we want the top bar style without the sticky behavior, for example in API documentation where there are also no tabs in the top bar.)

I think this makes sense and will help differentiate what is in use. I'll update!

katydecorah:
@colleenmcginnis this is ready again, overview:

- Created `Topbar` component which is the styling of the bar. I added a `limiter` wrapper inside, since this appears in every current TopbarSticker implementation right now.
- Reference Topbar in TopbarSticker
- Updated a few other testcases that use TopbarSticker to use Topbar.

katydecorah:
(p.s. this is no rush)

PR #276, opened 2020-05-07
Reconfigure `Video` options to optimize for user's settings around autoplay

### LINKED ISSUE #275: If user has video autoplay disabled, `Video` looks like an image (no controls to play)
**Describe the bug**

**To Reproduce**
1. In Firefox, go to `about:preferences`
2. Search for "autoplay"
3. Click Autoplay "Settings..."
4. Select "Block Audio and Video" and "Save changes".
5. Visit https://docs.mapbox.com/android/maps/examples/fit-camera-in-bounding-box/

The video does no autoplay. The only way to play the video is to right-click and select Play, but it's not clear to the user that it's a video.

**Expected behavior**

The user should have access to video controls. Perhaps we can determine if the user has autoplay disabled and then show a play button over the video.

katydecorah:
**This branch merges into #431** 

This PR makes some changes to how we handle autoplay, loop, and the displaying of controls in the `Video` component. Previously we handled users who prefer reduce motion by disabling loop, autoplay, and showing the controls. We did not take into account browser settings that would disable autoplay, so this PR will. 

The goal is to give preference to the user's settings first:

- Regardless of settings, on hover of the video container the controls will appear.
- Autoplay and loop settings are turned off by default. If the user doesn't prefer reduced motion, then we will set autoplay and loop to true (unless the those props are false as passed through to the component.)
- If the video is not playing, show a play button icon over the video. This will help indicate to users who prefer reduced motion and users who have disabled autoplay in their browsers that the content is a video. Once they click the button the video will play.

Fixes #275 

## Screenshots

image
> Play button appears for users who prefer reduced motion or have disabled autoplay

## How to test

Interact with /Video or /Phone test cases:
  - [ ] Try with your usual settings
  - [ ] Try with [enabled reduce motion](https://support.apple.com/guide/mac-help/reduce-screen-motion-mchlc03f57a1/mac)
  - [ ] Try with [disabled autoplay in your browser](https://support.mozilla.org/en-US/kb/block-autoplay )

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `7.0.0`.


src/components/video/video.js
```diff
@@ -1,45 +1,127 @@
...
+            <VimeoPlayImage />
+          </button>
+        )}
         <video
           {...videoProps}
-          muted
           width="100%"
           className="block mx-auto"
-          src={this.props.src}
+          src={`${src}#t=0.1`} // the hash enables preview on iOS
```
katydecorah, line 88: When testing on iOS, I noticed that when auto play is disabled, then the video will not show a preview (just a blank background with the play button). 

> 

I learned that iOS does this to prevent downloading the video until it's played. We have two options to generate the preview:

1. Use this hack, which tells the video player to advance the video timeline by 0.1 seconds, which will make iOS video show the preview.
2. Require that the video include an image, which the video component will use as the preview until the video starts playing/is downloaded.

Option 2 is the best practice, but it would require significant energy in rollout since we could need to generate a thumbnail image for every video (I'm thinking about android/ios examples which use a lot of video).

## What should we do?

1. Take an in between step to get us closer to the best practice: We enable this "hack" temporarily. We add a preview image option to Video so that moving forward, we will include a preview image with the video. We create an issue in documentation to quest the task of adding preview images to videos.

2. Stick with the hack. Since the blank video preview will only be seen by some iOS users who have autoplay disabled - this is likely a small fraction of our users.

3. Don't do the hack, do not add preview images (but add an optional preview image prop to make the component future friendly), and instead set a background color (or figure out a very generic preview image) to the component define the video area to prevent it from looking blank.

4. Something else?

@colleenmcginnis @danswick any opinions? I'm leaning toward number 3, since this issue likely effects very few users and while I strive for best practices, adding preview images to all videos would be a significant undertaking.

colleenmcginnis: @katydecorah I'm good with option 3! I think just a background color (that has good contrast with the play button) would be enough to suggest that there will be a video there even if it's not ready at the start.

danswick: Agreed! I think option 3 is a good place to start and we can revisit if it seems to be causing issues. Also agreed that just a background color is probably totally sufficient 👍

katydecorah: Thanks for the feedback! I'm going to remove the hack and add a background to this component.

src/components/video/video.js
```diff
@@ -1,45 +1,131 @@
...
+            Your browser doesn't support HTML5 video. Open{' '}
+            <a href={this.props.src}>link to the video</a>.
           </p>
         </video>
       </div>
     );
   }
 }
 
+// default props are only set if the user does not prefer reduced motion
```
danswick, line 107: Great feature!

src/components/video/video.js
```diff
@@ -1,45 +1,131 @@
...
+Video.defaultProps = {
+  autoplay: true,
```
danswick, line 109: Should autoplay be `false` by default?

danswick (commented):
This seems like a really smartly designed video component, @katydecorah! I was a little confused by autoplay and left a comment inline. When I tested locally, the basic example was autoplaying by default. Looking at `video.js`, that seems like the expected behavior, but the code comments and changelog entry suggest it shouldn't be. Am I misunderstanding that test case?

src/components/video/video.js
katydecorah: @danswick I can help improve the wording, but here's the gist:

1. The `state` by default disables autoplay. We assume that the user will prefer reduced motion.
2. Then in componentDidMount, we check the users's preferences:
    - If the user prefers reduced motion, the component will not use default props and will continue to disable auto play.
    - If the user does not prefer reduced motion, then the component will use default props to determine if the video should be autoplayed.

src/components/video/video.js
katydecorah, line 107:
```suggestion
// Video will use props if the user does not prefer reduced motion, otherwise it will default to the values in the state.
```

src/components/video/video.js
```diff
@@ -1,45 +1,131 @@
+/* eslint-disable jsx-a11y/media-has-caption */
 import React from 'react';
 import PropTypes from 'prop-types';
+import { VimeoPlayImage } from '../related-page/vimeo';
```
katydecorah, line 5:
```suggestion
// The state defines the user's preferences for the video settings (autoplay and loop).
// The props for Video will only be taken into account if the user does not prefer reduced motion.
// In `componentDidMount`, Video will determine the user's settings. 
//  - If the user prefers reduced motion, then Video will use the `state` which disables autoplay and loop.
//  - If the user does not prefer reduced motion, then the props will be taken into account.
```

danswick (approved):
Thanks for tweaking the comment language, @katydecorah! Looks great to me 🥞

PR #292, opened 2020-06-14
Set peerDependencies to keep frontend libraries in sync

### LINKED ISSUE #285: Make sentry/browser a peer dependency
We should make sure that the version in dr-ui matches the version in the site it's installed in.

Ref: https://github.com/mapbox/mapbox-gl-js-docs/pull/343

katydecorah:
This PR updates sentry/browser and sets peerDependency. Fixes #285 

To help keep sites in sync, this PR also:
- Updates react and react-dom and sets as peerDependencies.
- Adds mr-ui and mbx-assembly as peerDependencies.


katydecorah:
I think we should also make react and react-dom peerDependencies so we can have a consistent version across repos.

PR #313, opened 2020-07-09
Add Sentry to forward-event client

katydecorah:
So continues our quest to capture forward-event issues.

A few months ago we added Sentry to forward-event repo, but I think that we also need to add it to our client that invokes forward-event.

This will give us:
- Exact page the failed forward-event request happened on.
- The event payload

## How to test

1. Pull down this branch && `npm ci`.
2. Replace `xhr.send(JSON.stringify(event));` with `xhr.send('test'));` This will trigger a 4xx error.
3. Run `npm start` and visit the /Feedback test case.
4. Submit feedback for the first feedback component.
5. Check to see that a 400 HTTP error event came through at `docs-forward-event` Sentry Project.

### DIFF
--- a/src/components/feedback/forward-event.js
+++ b/src/components/feedback/forward-event.js
@@ -1,3 +1,9 @@
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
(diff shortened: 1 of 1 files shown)


katydecorah:
No longer needed

PR #331, opened 2020-09-21
[info arch] Create `Breadcrumb` component

katydecorah:
**This PR merges into `info-arch-main`.**

This PR will introduce a new component `Breadcrumb` that will be included in `PageLayout`.

## How to test

- [ ] View /Breadcrumb tests cases app (npm start)
- [ ] View component used on /guides on catalog site (npm run start-docs)

## QA checklist

- [x] Create a PR in a site repo, copy the component, and test it. Push to staging and let the reviewer know they can also test the component there. *It's being used on the catalog site, run `npm run start-docs`*


src/components/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,84 @@
...
+      path: PropTypes.string.isRequired
+    }),
+    PropTypes.bool
+  ]),
+  /** Title of the site and homepage path */
+  site: PropTypes.shape({
+    title: PropTypes.string.isRequired,
+    path: PropTypes.string.isRequired
+  }),
+  /** Subsite title and homepage path (if using multi-structured layout) */
```
danswick, line 62: `subsite` and `section` just dictate the order of the breadcrumbs, right? Are there any scenarios where we might want multi-level breadcrumbs for something that doesn't have a true subsite?

src/components/breadcrumb/examples/basic.js
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react';
+import Breadcrumb from '../breadcrumb';
+
+export default class Basic extends React.Component {
+  render() {
+    return (
+      <Breadcrumb
+        currentPage={{
+          title: 'Display a map',
+          path: '/mapbox-gl-js/examples/display-a-map/'
```
danswick, line 14: Do you think it's important that the current page's path is consistent with the `site/subsite/section` path? I was wondering if we might want to add some kind of validation utility but maybe that would be too restrictive.

src/components/breadcrumb/__tests__/utils.test.js
```diff
@@ -0,0 +1,25 @@
...
+          path: '/api/'
+        },
+        {
+          title: 'API',
+          path: '/api/'
+        }
+      ])
+    ).toEqual([
+      { path: '/api/maps/', title: 'Maps service' },
+      { path: '/api/', title: 'Overview' }
```
danswick, line 22: If there are two titles with the same path, will this utility always return the first one? I wonder if we might want to throw instead to alert contributors that they have included two of the same path. Or is there another scenario you have in mind where we would expect two of the same path to make it here?

src/components/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,84 @@
...
+      ...(domain !== false ? [domain] : []),
+      ...(site ? [site] : []),
+      ...(subsite ? [subsite] : []),
+      ...(section ? [section] : []),
+      ...(currentPage ? [currentPage] : [])
+    ]);
+
+    return createLinks.length > 1 ? (
+      <div
+        className="dr-ui--breadcrumb none block-mm px24 pt12"
```
danswick, line 31: We're ok to set these padding values here because `<Breadcrumb>` is a layout component and is expected to be outside of the main `Content`, right?

danswick (commented):
This is look really great! I ran the test cases app and the catalog locally and everything looks and behaves exactly like I would expect. I left a few higher level questions inline for us to chat about.

src/components/breadcrumb/__tests__/breadcrumb-test-cases.js
```diff
@@ -0,0 +1,140 @@
...
+        title: 'Camera',
+        path: '/android/maps/overview/camera/'
+      }}
+      site={{
+        title: 'Android',
+        path: '/android/'
+      }}
+      subsite={{
+        title: 'Maps SDK',
+        path: '/android/maps/overview/'
```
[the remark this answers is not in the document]
katydecorah: https://docs.mapbox.com/android/maps/ redirects to https://docs.mapbox.com/android/maps/overview/

src/components/breadcrumb/breadcrumb.js
katydecorah: > subsite and section just dictate the order of the breadcrumbs, right? 

Correct

> Are there any scenarios where we might want multi-level breadcrumbs for something that doesn't have a true subsite?

I'm unsure of a scenario right now, but if we find one we can add a test case and fix to cover it.

src/components/breadcrumb/breadcrumb.js
katydecorah: This padding was designed to fit nicely into PageLayout (as it appears now). If we think we'd use this component elsewhere, then we would probably want to remove the padding here and add a wrapper div (or something) to add the necessary padding when this component is referenced in PageLayout.

src/components/breadcrumb/examples/basic.js
katydecorah: > Do you think it's important that the current page's path is consistent with the site/subsite/section path?

Not sure I'm following, can you provide an example of what this would look like?

src/components/breadcrumb/__tests__/utils.test.js
katydecorah: If there are two titles with the same path, the utility should return the first one. It should be a rare case, but when I was making test cases it did show up. Ideally, the logic should be builtin to PageLayout so the contributor shouldn't have to worry about configuring this component at all. It may take some trial and error during the testing phase to identify any one offs and then create tests cases, and update the logic.

My only concern with throwing an alert is that unless it breaks the build, then it can be really easy for the contributor to miss the alert. It could be worth it to think this concept of alerts.

src/components/breadcrumb/breadcrumb.js
katydecorah: 🦆 I think we should remove the padding to keep this component independent. I'll update!

katydecorah: 🦆 Or maybe, a theme prop to be able to pass classnames to the container!

katydecorah:
@danswick Thank you for your comments! I made two changes that I think will help make this component more flexible:

1. Removed the classes from the containing div from the the component and replaced with `themeWrapper` prop to be able to pass Assembly classes.
2. Removed `site`, `subsite`, `section`, and `currentPage` props and replaced with `links` array. The breadcrumb will now use the links arrays to generate each crumb in the order provided.

One thing I'm wondering: should we remove the `createUniqueCrumbs` from the Breadcrumb component? This would provide the most amount of flexibility to the component as it would display the links exactly as described by the `links` array. We could then move the `createUniqueCrumbs` function into PageLayout to help make sure links are unique.

katydecorah:
@danswick this is ready again! Per chat today, I moved `createUniqueCrumbs` util function out of Breadcrumb and into PageLayout. I also removed the `domain` prop from Breadcrumb, as that can be added to the new `links` array.

src/components/breadcrumb/breadcrumb.js
danswick: The new array setup looks really great to me!

src/components/breadcrumb/examples/basic.js
danswick: Sorry, that wasn't worded very well. I was thinking about a scenario like: 

`All pages > sub-page 1 > sub-page 2 > current page`

Where the pages' paths are not hierarchical: 

```
All pages: site.com
Sub-page 1: site.com/products
Sub-page 2: site.com/product/widget-api
Current page: site.com/product/widget-api/exampes
```

The above example is kind of like how you can go to `github.com/mapbox/dr-ui/pulls`, but PR 123 is at `github.com/mapbox/dr-ui/pull/123`. 

I can't think of anything in our subdomain like this and can't imagine we would really need to verify if a particular path follows some predetermined format. I guess what I was getting at was asking if you thought we would need any kind of validation that the `current site` path included all of its parent path. This 🦆 session has gone on too long now 😂

src/components/breadcrumb/breadcrumb.js
danswick: It's beautiful!

danswick (approved):
Your changes all look really great to me!

src/components/breadcrumb/examples/basic.js
katydecorah: Lol! Thanks for clarifying. I see what you're saying and our new systems of passing an array of links should support this!

Issue #332, opened 2020-09-21
[info arch] Redesign PageLayout


katydecorah:
As part of doc's information architecture work, we'll be updating `PageLayout` to improve the navigational systems of all sites:

- [ ] Remove TopbarSticker for PageLayout.
- [ ] Update NavigationAccordion 
  - Move navigation to the left sidebar, we can make design adjustments to NavigationAccordion to act as the new main navigation for the site.
  - Consider: Make NavigationAccordion items more distinct #280
  - Consider: Reduce whitespace in NavigationAccordion #279
- [ ] Add Search component to left sidebar (if we’re unable to add it to the docs-page-shell during this first iteration).
- [ ] Adjust max-width of the `limiter` class on docs pages to make more room for the new aside feature ref (#342)

Start: 10/1
Request review: 10/16
Merge: 10/23

katydecorah:
I added another task related to PageLayouts:

> Adjust max-width of the `limiter` class on docs pages to make more room for the new aside feature ref (#342)

PR #342, opened 2020-10-16
[info arch] Adds OnThisPage component and aside feature to PageLayout

### LINKED ISSUE #341: [info arch] Create sticky Aside for PageLayout
- [x] Move headings to sticky aside.
- [x] Decide: what heading levels should we show? H2 and H3 (display h3s only when active on h2 section)? Use scrollspy to indicate which heading you're on?
- [x] Decide: what happens on smaller devices? Should it go inline on smaller devices and then stick to the side on larger devices?
- [x] Research: can we remove the plugin to markup pages and use scroll observer to find h2/h3 sections?
- [x] Update Feedback component.
  - Make Feedback component adjustments as needed to fit in the new location

katydecorah:
This PR merges into #334 

- Create `OnThisPage` component that displays headings on the page and performs scroll spy to indicate where you are on the page.
  - Replaces react-scrollspy with a local script to detect when a heading is in view.
  - Deletes `@mapbox/dr-ui/plugins/create-sections` plugin as it's no longer needed.
- Update `PageLayout` component.
  - Add aside feature which includes the `OnThisPage` and `Feedback` components. On larger devices, the aside is stuck to the right-side of the screen. On smaller devices, `OnThisPage` moves inline below the page's title and `Feedback` will appear at the bottom of the page.
- Update `Feedback` component.
  - Removed background color and use `AsideHeading` component to style the component's heading.
- Update `docs-prose.css`.
  - You can now use `.unprose` class on `h2` elements to remove the styling.

Note: The dr-ui component home page doesn't have a left sidebar right now. It will once #332 lands.

 In addition to the details in the changelog, I wanted to point out a few decisions and get your thoughts:

- I decided not to use `Sticky` component (react-stickynode) for the aside feature in PageLayout. I decided to go with a  less expensive option, CSS (`position: sticky`), however this means that the aside feature will not stick on IE 11 and will instead by static. I think this is ok and considered a progressive enhancement as the core functionality of the component of providing anchor links is still attainable.

- The new `OnThisPage` component requires that all headings that will be used as links have the `anchor` class. This will help us to better identify and track via the scroll spy functionality. This should not be a problem for most pages, where this PR updates the add-links-to-headings plugin to add the class to headings automatically. This also allows us to remove (which this PR does) the create-sections plugin which added a lot of divs to sectionize headings for the purpose of scroll spy. Overall, I think this will be an improvement. But during roll out, we'll need to be mindful of pages that use html rather than markdown (like Mapbox GL JS) to make sure we add the `anchor` class to each h2 and h3 element that is expected to show up in the OnThisPage component.

- This PR also includes a bunch of small/fiddly updates to the catalog page to enable the Feedback component on all pages to make testing easier.

## To do

- [x] Fix tests
- [x] Fix aside on catalog page (it scrolls with the page)
- [x] Add task to #332 to increase max-width of `limiter` class in `PageLayout` to give the content and new feature enough space.

## How to test

- View the OnThisPage test cases page (npm start)
- View the catalog site (npm run start-docs) -- Note the homepage of the catalog site doesn't currently have a left sidebar, but that will be changed when the PageLayout updates are added.

## QA checklist

Fixes #341


src/components/card-container/card-container.js
```diff
@@ -26,7 +26,7 @@ class CardContainer extends React.PureComponent {
     return (
       <div>
         <a href={props.path} className="unprose mb18 block color-blue-on-hover">
-          <h2 className="txt-bold" id={categoryID}>
+          <h2 className="txt-bold unprose" id={categoryID}>
```
katydecorah, line 29: Note: Due to the arranging of the PageLayout content component, needed to add `unprose` here to prevent these headings from getting styled.

src/components/on-this-page/__tests__/__snapshots__/on-this-page.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`OnThisPage Basic renders as expected 1`] = `undefined`;
```
katydecorah, line 3: Need to fix this!

src/components/on-this-page/helpers.js
```diff
@@ -0,0 +1,108 @@
...
+          ...(sub.icon && { icon: sub.icon })
+        }))
+      });
+    }
+    return arr;
+  }, []);
+}
+
+// finds the active header anchor
+export function getActiveHeaderAnchor(topOffset) {
```
katydecorah, line 37: This function was heavily borrowed from https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/hooks/useTOCHighlight.ts

danswick (approved):
This is so good! I tested thoroughly locally and left a few comments (mostly just clarifying) inline.

Regarding the prompts you left: 
* _I decided not to use Sticky component..._ This seems like a great decision to me. The IE trade-off seems perfectly justifiable to me as well.
* _The new `OnThisPage` component requires that all headings that will be used as links have the anchor class..._ This also seems like a justifiable decision. Most of this component's users will be working almost exclusively in markdown anyway. If it becomes a big issue, we could consider a custom plugin, but I feel pretty confident we won't need to.

My only concern when looking at the live page is that 1200px might be a bit of a high threshold to sacrifice both a persistent sidebar as well as second-level headings. Right now, anything that takes up less than half the width of my external monitor gets the collapsed version. Maybe we can check back in when the page layout is finalized to see how much real estate we can buy back and see what feels feasible?

katydecorah:
> My only concern when looking at the live page is that 1200px might be a bit of a high threshold to sacrifice both a persistent sidebar as well as second-level headings. Right now, anything that takes up less than half the width of my external monitor gets the collapsed version. Maybe we can check back in when the page layout is finalized to see how much real estate we can buy back and see what feels feasible?

@danswick Right, I added a task to https://github.com/mapbox/dr-ui/issues/332 to adjust the max-width of the limiter to afford space for the aside

src/components/on-this-page/on-this-page.js
```diff
@@ -0,0 +1,146 @@
...
+  headings: PropTypes.arrayOf(
+    PropTypes.shape({
+      text: PropTypes.string.isRequired,
+      slug: PropTypes.string.isRequired,
+      children: PropTypes.arrayOf(
+        PropTypes.shape({
+          text: PropTypes.string.isRequired,
+          slug: PropTypes.string.isRequired,
+          icon: PropTypes.string
+        })
```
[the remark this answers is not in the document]
katydecorah: I think we are committed to h2 and h3. I think anything beyond h3 is too granular.

src/components/on-this-page/on-this-page.js
```diff
@@ -0,0 +1,146 @@
...
+    // if activeHeaderAnchor has an id, set it as the lastActiveLink
+    if (activeHeaderAnchor && activeHeaderAnchor.id) {
+      this.setState({
+        lastActiveLink: activeHeaderAnchor.id
+      });
+    }
+  };
+
+  // if the h2 or a (sub) h3 id matches `lastActiveLink`, mark this section as active
+  findActiveSection = (heading) => {
```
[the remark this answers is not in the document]
katydecorah: That's a good idea! I'll update

PR #267, opened 2020-04-01
Replace Select in `NavigationDropdown` with an accessible menu

katydecorah:
This PR:
- Replaces the Select element in NavigationDropdown with an accessible menu: follows recommendations for WAI-ARIA practices and fully keyboard functional.
- Reduces padding on wrapping container in `NavigationAccordion`.

## How to test

Try out the `NavigationDropdown` and `NavigationAccordion` (mobile sizes) test cases.

## QA checklist

- [x] No errors logged to console.
- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `#.#.#`.
- [x] Component is accessible at mobile-size viewport.
- [x] Component is accessible at tablet-size viewport.
- [x] Component is accessible at laptop-size viewport.
- [x] Component is accessible at big-monitor-size viewport.
- [x] Create a PR in a site repo, copy the component, and test it. Push to staging and let the reviewer know they can also test the component there.

Open the test cases app locally on:

- [x] Chrome, no errors logged to console.
- [x] Firefox, no errors logged to console.
- [x] Safari, no errors logged to console.
- [x] Edge, no errors logged to console.
- [x] IE11, no errors logged to console.
- [x] Mobile Safari, no errors logged to console.
- [x] Android Chrome, no errors logged to console.

Fixes #248

PR #344, opened 2020-10-27
[Info arch] PageLayout major structural changes

danswick:
Fixes https://github.com/mapbox/dr-ui/issues/332

This PR: 

- [x] Removes topbar sticker.
- [x] Removes the accordion page layout.
- [x] Adds a new "page" page layout with nav comprised of all the major site sections and their children instead of headings.
- [x] Updates `NavigationAccordion` component for page layout.

Quick preview:

image

Each section of the site, previously known as a "Tab," is now an item in the left sidebar. Any site sections with multiple children get a drop-down listing each of those children.

The headings from the previous version of the `NavigationAccordion` are now in the right-hand aside.

## How to test

1. Pull down this branch and `npm install`. 
2. Start the Dr. UI catalog site with `npm start-docs`.

## Open questions and follow-up tasks

1. How would we like to handle this sidebar on mobile? Maybe it could turn into top tabs (the irony) for small viewports. Right now it does nothing.
2. It does not currently support tags. I was having some trouble understanding how to configure tags and unfortunately ran out of time.
3. There are some static build failures causing tests to fail. I suspect I've goofed something in the catalog docs, but am having trouble tracking it down.

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `6.2.0`: `29`.

- [ ] Edge, no errors logged to console. _None of the pages of the test app load in IE_
- [ ] Mobile Safari, no errors logged to console. _I don't think browserstack supports localhost on i-devices?_


docs/src/pages/page/grizzlies.md
```diff
@@ -0,0 +1,13 @@
+---
```
danswick, line 1: Just some temporary pages while testing.

src/components/navigation-page/navigation-page.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
```
danswick, line 1: This navigation scheme should be used for each layout type now.

src/components/page-layout/components/page-sidebar.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
danswick, line 1: The only purpose of this component is to pass the `parentPage` prop to the `NavigationPage` component.

src/components/page-layout/components/sidebar.js
```diff
@@ -55,45 +54,19 @@ export default class Sidebar extends React.Component {
...
       children,
       parentPath,
-      headings,
       topics,
       layoutConfig
     } = this.props;
 
-    const { accordion } = navigation;
-
+    /* Should we support some of these other features as well?
```
danswick, line 61: @katydecorah I wasn't sure if we should consider passing some of these configuration options in to the new sidebar.

danswick:
@katydecorah I'm hitting a bit of a stumbling block related to parent-child pages. There are a couple of things in particular: 

1. For a section like [Guides](https://mapbox.github.io/dr-ui/guides/), there's a bit of awkwardness where the "parent" page is also one of the children. This issue is present (IMO) in the current layout, but is not as obvious or weird (see below). It is more obvious and awkward in the new layout.

  image

  _Not super awk, but still duplicative._

  image

  _Awk double `Guides`_

2. Because of the way this branch is looking at parent-child relationships between pages, the root of `dr-ui`'s docs site just lists every page with `layout: page` as a sub-item. We could use the "hide sub-items" option here, but that seems less than ideal since this is likely to be an issue with every page's root, at least with the current setup. Do you have any ideas for how we can keep the root page from listing every other page on the site? Maybe we need a `findImmediateChildren` and/or `findGrandChildren` [helper function](https://github.com/mapbox/dr-ui/blob/ed5f0989269d3148483df63cd9488e0a5e6f80e6/src/helpers/batfish/navigation.js#L46). Very open to ideas! 

  image

  _Quite awk._

src/components/navigation-page/navigation-page.js
katydecorah: Rather than create a new component, can we replace NavigationAccordion? We have no use for NavigationAccordion otherwise.

katydecorah:
> For a section like Guides, there's a bit of awkwardness where the "parent" page is also one of the children. This issue is present (IMO) in the current layout, but is not as obvious or weird (see below). It is more obvious and awkward in the new layout.

Can we filter out parent pages so they aren't included as subpages? 

> Because of the way this branch is looking at parent-child relationships between pages, the root of dr-ui's docs site just lists every page with layout: page as a sub-item. We could use the "hide sub-items" option here, but that seems less than ideal since this is likely to be an issue with every page's root, at least with the current setup. Do you have any ideas for how we can keep the root page from listing every other page on the site? Maybe we need a findImmediateChildren and/or findGrandChildren helper function. Very open to ideas!

I think this is moreso a quirk of how this Catalog site was built and not likely a file pattern used by other sites. To fix the hierarchy here, we can:

1. Rename `docs/src/pages/index.md` -> `docs/src/pages/overview/index.md` 
2. Create `docs/src/pages/index.js` which redirects to the overview page.

src/components/navigation-page/navigation-page.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+  }
+
+  componentDidMount() {
+    const { location, parentPage } = this.props;
+
+    this.setState({
+      activeItem: location.pathname,
+      activeParent: parentPage
+    });
```
katydecorah, line 20: Since this component isn't interactive, I don't think we need to use state. Instead we should be able to set the component based on the props passed.

src/components/page-layout/components/sidebar.js
katydecorah: I don't think so. The new sidebar should be pretty opinionated to help keep us consistent. It's likely we'll open up more configuration options to the PageLayout aside, though.

src/components/page-layout/components/page-sidebar.js
```diff
@@ -0,0 +1,37 @@
...
+    return pageIsTab
+      ? location.pathname
+      : findParentPath(navigation, location.pathname);
+  }
+
+  render() {
+    const { navigation, location } = this.props;
+    
+    return (
+      <NavigationPage
```
katydecorah, line 25: We'll also need to include the `ProductMenu` which was in the TopBarSticker to identify the site's name.

src/components/navigation-page/navigation-page.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  renderHeader(href, label) {
```
katydecorah, line 23: I mocked up some suggestions on styling. I think your structure is really good, but we can play with the spacing and color to improve the hierarchy structure between main navigation items and subpages:

image

src/components/page-layout/layout.config.js
```diff
@@ -5,20 +5,17 @@ export default {
   },
   accordion: {
     sidebar: 'accordion', // NavigationAccordion sidebar
```
katydecorah, line 7: I think we can remove the `sidebar` configuration option. In my opinion, there should only be one sidebar with no customizations (since this is the site's main navigation and we didn't have offer customization with NavTabs). We'll move customization options to the PageLayout aside.

src/components/page-layout/page-layout.js
```diff
@@ -101,9 +76,6 @@ export default class PageLayout extends React.Component {
     return (
       <ErrorBoundary>
         {!noShellHeaderBuffer && <div className="shell-header-buffer" />}
-        <ErrorBoundary>
-          {this.renderTopbar(switchedNavigation, parentPath)}
-        </ErrorBoundary>
         <div className="limiter">
```
katydecorah, line 79: We may want to create our own limiter class -- perhaps one with a max-width of 1800 pixels.

src/components/page-layout/layout.config.js
katydecorah: We can save making these updates for a later PR.

src/components/navigation-page/navigation-page.js
danswick: Thanks for this, Katy! I took your suggestions and run with them a bit: 

1. The main "Tab" section is bubbled now. I didn't love the fully round highlights (to me, they look a bit too much like buttons), so I tried only rounding the left side and I like it! It feels a bit like a "Tab," go figure.
2. Chevron-y icons are in place!
3. I removed a lot of whitespace, which I think makes the hierarchy a bit clearer at a glance.

image

src/components/page-layout/page-layout.js
danswick: I spent a few minutes testing this out and made a horrible mess of things, so decided to leave it for now.

src/components/page-layout/page-layout.js
```diff
@@ -15,43 +14,19 @@ import layoutConfig from './layout.config.js';
...
-  };
-
   // render the page's sidebar
   renderSidebar = (config, switchedNavigation, parentPath) => {
-    const { customSidebar, headings, topBarSticker } = this.props;
+    const { customSidebar, constants } = this.props;
     return (
       config.sidebar !== 'none' && (
-        <div className={`col col--4-mm col--12 ${config.sidebarTheme}`}>
+        <div className={`col col--3-mm col--12 ${config.sidebarTheme}`}>
```
danswick, line 22: I decreased the columns for the sidebar now that it's a bit simpler and added a column to the content section below.

katydecorah:
This is great work @danswick 

I hopped in with some tweaks after testing locally on a few docs sites. Here's an overview of what I worked on (I know this looks like a lot, but it accounts for some quirks from all our sites which tend to be small but add up fast 😅 please let me know if you want to discuss any of these changes):

- Remove `txt-fancy` from the PageLayout aside headers to help make them feel like secondary headings.
- Made some slight styling adjustments to the NavigationAccordion to help increase visual hierarchy and build distinctions between each item.
- Replace react-stickynode component with CSS sticky prop (like we did with aside).
- Made some adjustments to PageLayout Breadcrumb implementation based on testing on sites. (Remove "for Android/iOS" from site titles to reduce repetition.)
- For the PageLayout content area, replace use of `grid` (to space the sidebar and content area) with flexbox. This allows us to set the sidebar width to prevent it taking up too much space or not enough depending on the device size).
- Add frontMatter options to `hideSidebar` and `hideBreadcrumbs` which are needed for the Help homepage. And `hideSupages` for /help Tutorial, Troubleshooting, Glossary pages to prevent all the pages from appearing in the sidebar.
- Create `limiter--wide` class to set a width of 1800px on larger screens.
- Remove additional padding that was needed on headings to clear the TopbarSticker.
- In the navigation Batfish helper, move "accordion" data under each navTab item as `pages` to bundle navigation dataset all together for easier traversing.


## Things to do/discuss

- [ ] **Update docs-page-shell** to use our new limiter class and remove "overflow: hidden" on container which prevents CSS `sticky` from working. Ref https://github.com/mapbox/docs-page-shell/pull/88
- [ ] 🗣 **How will we handle NavigationAccordion content on mobile?** A couple ideas so far:
   1. I started reworking NavigationDropdown, we could use that for the mobile experience #267 (the original NavigationAccordion component used NavigationDropdown on mobile)
   2. We could make the chevron icon toggle the open / closed state of subpages (github docs does this). On mobile, we can collapse the toggle by default for active parent pages with subpages.
- [ ] 🗣  **On overview pages on smaller screens, OnThisPage component goes above the OverviewHeader.** A couple ideas so far:
   1. We move OverviewHeader options to frontmatter and load OverviewHeader above OnThisPage in the PageLayout component.
   2. We remove the required title prop from OverviewHeader so that the page will load an h1 (markdown h1's get place before OnThisPage feature). Only problem: we'd likely need to add an `altTitle` property as the title of many overview pages does not match the title used in the Overviewheader.

docs/src/pages/guides/page-layout.md
```diff
@@ -151,27 +139,62 @@ Things to consider:
...
+  })
+}
+```
+
+This technique may also be used by multi-structured sites, although you'll need to add logic to help determine which version constant to use.
+
+## Custom aside
 
-Sites like Mapbox GL JS require custom sidebars since the data is derived from multiple sources. Similarly, the [Dr. UI component's page](/dr-ui/) also uses a custom sidebar, see the example below for how to conditionally display a custom sidebar:
+Sites like Mapbox GL JS require custom asides since the data is derived from multiple sources. Similarly, the [Dr. UI component's page](/dr-ui/) also uses a custom aside, see the example below for how to conditionally display a custom sidebar:
```
danswick, line 187:
```suggestion
Sites like Mapbox GL JS require custom asides since the data is derived from multiple sources. Similarly, the [Dr. UI components page](/dr-ui/) also uses a custom aside, see the example below for how to conditionally display a custom sidebar:
```

src/components/overview-header/overview-header.js
```diff
@@ -109,30 +102,34 @@ class OverviewHeader extends React.PureComponent {
...
     return (
-      <div className="scroll-hidden border-b border--gray-light prose mb24">
+      <div className="dr-ui--overview-header border-b border--darken10 prose mb24 pr60-mxl">
         <h1 className="mb6 txt-fancy">
           {props.title}
           {props.tag && this.buildTag(props)}
         </h1>
-        <div className="relative">
-          <div className="pr12-ml mr240-ml mr0">
+        <div className="flex-parent">
```
danswick, line 110: What do you think about moving the `<h1>` inside the left-hand, text area? I noticed there might be some spacing funkiness between the image on the right and the text on the left: 

image

👇 

image

danswick (commented):
Thank you so much for all of this great work, @katydecorah! I went through all of the changes and everything looks really thoughtful and solid to me. I left one suggestion for a small copyedit (already present) and one related to spacing in the header image.

src/components/overview-header/overview-header.js
katydecorah: Updated!

Issue #333, opened 2020-09-21
[info arch] Create Filter component (or update ExamplesPage to add filters)


katydecorah:
We'll be adding a set of filters to `exampleIndex` layout pages to improve findability of examples.

- [ ] Decide: Should this feature be a new component or should we redesign the ExamplesPage component?
- [ ] When the data exists, add toggles for: language, level, product, topic/s, videos.
- [ ] Filters can be controlled by URL parameters.
- [ ] Decide: How should Batfish helper “topics” work moving forward? This dataset was created to help organize pages by topic, but the filter feature will provide much more control.

---

Build out suggestions:

1. Rename or replace Batfish helper "topics" -- For documentation about this function, see: https://mapbox.github.io/dr-ui/guides/batfish-helpers/#topics
    - We will need to have an array of sub pages that will appear on this page. It could be helpful to sort them by topic by default, this could help make sure the "Getting Started" examples appear first.
    - Could be helpful to generate unique arrays for all the datasets we want to filter based on the the sub pages (else we do this client-side)
2. Build out filter functionality in `/src/components/page-layout/components/layouts/example/content.js`.
    - This could include a sub component that displays the title for the filter group and then buttons or toggle for each option. This sub component would (likely) only display if there is more than one options for the filter.
    - The only filter feature that may be different is the "Videos only" toggle as seen on Help examples. So this may need to have a different treatment.

---

Start: 10/19
Request review: 10/23
Merge: 10/30

PR #343, opened 2020-10-27
[Info arch] Page layout main

danswick:
fixes https://github.com/mapbox/dr-ui/issues/332

This PR will be home to the PageLayout redesign "feature branch." The redesign will encompass a number of smaller changes that will each be merged into this branch before merging into the main info arch branch.

## QA checklist

- [ ] No errors logged to console.
- [ ] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `#.#.#`.
- [ ] Component is accessible at mobile-size viewport.
- [ ] Component is accessible at tablet-size viewport.
- [ ] Component is accessible at laptop-size viewport.
- [ ] Component is accessible at big-monitor-size viewport.
- [ ] Create a PR in a site repo, copy the component, and test it. Push to staging and let the reviewer know they can also test the component there.

Open the test cases app locally on:

- [ ] Chrome, no errors logged to console.
- [ ] Firefox, no errors logged to console.
- [ ] Safari, no errors logged to console.
- [ ] Edge, no errors logged to console.
- [ ] IE11, no errors logged to console.
- [ ] Mobile Safari, no errors logged to console.
- [ ] Android Chrome, no errors logged to console.

PR #345, opened 2020-11-02
[info arch] Add filters to exampleIndex layout

danswick:
Refs https://github.com/mapbox/dr-ui/issues/333

This PR will: 

1. Update batfish helpers to generate filter-by data for language, ~~product~~, level, and videos. These are in addition to the existing data for topics.
2. Add a filter component or build into examples layout.

This PR adds filter functionality to the `examplesLayout` and replaces the `topics` Batfish helper with the `filters` Batfish helper to power the new filtering feature. This PR does not propose creating new filter components, but if we find a need to reuse the elements used in PageLayout then we can move them into their own component.

## How to test

- View the catalog Examples page (`npm run start-docs` -> http://localhost:8080/dr-ui/examples/).
    - Test out each filter and test out query string params.

### Review discussion, notes

Some initial thoughts, happy to meet to discuss if that's easier:

- Curious what you think about the filters and how they function:
    - Right now you can only select one item from each filter category.
    - Should we only provide filters for topics and videos? Is levels and languages too much? Note: filters will only appear if there is more than one option available. So a site like Mapbox GL JS would not have a language filter because examples only use `JavaScript` and it wouldn't show a level filter because we haven't set that field on those pages.
   - Should we add filter by product? This PR doesn't include it yet as the options already feel robust. 
- How should the pages be ordered by default? Right now the pages are ordered by `order` if it exists in the frontmatter, and then alphabetically by title. This is a pattern we already use on subpages and would be easier than implementing a new system. This technique would also allow us to quickly surface certain pages to the top of the list.

## Screenshots

Demo of what the page would look like on Help:

> image

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `6.2.0`.
    -  Should be 98% due to #348 


docs/batfish.config.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path');
 const {
   buildNavigation,
-  buildTopics,
+  buildItems,
```
danswick, line 4: I generalized the `buildTopics` function to accept a new argument denoting which "filterable" to generate data for.

docs/batfish.config.js
```diff
@@ -40,7 +40,12 @@ module.exports = () => {
     },
     dataSelectors: {
       navigation: (data) => buildNavigation(siteBasePath, data),
-      topics: (data) => buildTopics(data, appendTopics),
+      topics: (data) =>
+        buildItems({ data, append: appendTopics, itemType: 'topics' }),
+      level: (data) => buildItems({ data, itemType: 'level' }),
+      language: (data) => buildItems({ data, itemType: 'language' }),
+      products: (data) => buildItems({ data, itemType: 'products' }),
+      videos: (data) => buildItems({ data, itemType: 'videos' }),
```
danswick, line 48: I was not able to fully investigate how we might want to build `append[Item]`s for each of the other "filterables."

docs/src/components/page-shell.js
```diff
@@ -4,6 +4,10 @@ import PageLayout from '../../../src/components/page-layout';
 
 import navigation from '@mapbox/batfish/data/navigation'; // eslint-disable-line
 import topics from '@mapbox/batfish/data/topics'; // eslint-disable-line
+import level from '@mapbox/batfish/data/level'; // eslint-disable-line
+import language from '@mapbox/batfish/data/language'; // eslint-disable-line
+import products from '@mapbox/batfish/data/products'; // eslint-disable-line
+import videos from '@mapbox/batfish/data/videos'; // eslint-disable-line
```
danswick, line 10: I could imagine bundling all of these together into a larger "filterables" object, but we could leave them like this for now and follow up later!

src/helpers/batfish/topics.js
```diff
@@ -23,99 +23,109 @@ function buildTopics(data, append, sortingArr) {
...
+  return uniqueItems.reduce((set, item) => {
     const subPages = pages.reduce((arr, page) => {
-      if (page.topics) {
-        if (page.topics.indexOf(topic) > -1) {
+      if (itemType === 'level' && typeof item === 'number') {
+        item = item.toString();
+      }
+      if (itemType === 'videos' && typeof item === 'boolean' && !!item) {
+        item = item.toString();
+      }
```
danswick, line 97: IMO, this is kind of a blunt way to handle `number` and `boolean` data.

src/helpers/batfish/topics.js
```diff
@@ -23,99 +23,109 @@ function buildTopics(data, append, sortingArr) {
...
+      if (page[itemType]) {
+        if (page[itemType].toString().indexOf(item) > -1) {
```
danswick, line 99: Again dealing with `number` and `boolean` types here. Maybe we want to handle this further up the chain or maybe it's ok!

danswick:
@katydecorah I got started here preparing filterable data for language, level, videos, and product.  Hopefully this is a useful starting point!

docs/src/components/page-shell.js
katydecorah: I agree, I think we should bundle these into a single object to avoid having to have added configuration in the page-shell and and batfish.config. I'll make some adjustments to the batfish helper to help bundle them.

katydecorah:
@colleenmcginnis this is ready for review, I updated the original comment with additional details to help during your review. Please let me know if you want to meet to discuss, as I think there are a few directions we can take filtering.

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -116,22 +112,25 @@ module.exports = () => {
...
 
-- The shape of topics is an object, where the top-level keys are pathnames for top-level pages that have subpages with `topics` or `topic`.
-  - Each object has a `topics`. It contains a unique list of topics, ordered by count of pages with that topic.
-    - Each topic has `pages`. It contains metadata for each page that has that topic.
+- The shape of filters is an object, where the top-level keys are pathnames for `examplesIndex` pages.
+  - Each object has a `topics` array, a unique list of topics.
+  - Each objects has `pages` array of all sub pages to display. The `pages` array is ordered by the `order` frontMatter property and then alphabetically by title.
+  - If available, the object may have a `levels` array, a unique list of levels options.
+  - If available, the object may have a `languages` array, a unique list of language options.
+  - If available, the object may have a `videos` array, an array of all pages with `video: true` in the frontMatter.
```
katydecorah, line 133: Note: the next section "Append topics helper" will be deleted in #347 so I didn't update old references to topics here.

src/components/page-layout/components/layouts/example/appropriate-image.js
```diff
@@ -1,12 +0,0 @@
-import { scopeAppropriateImage } from '@mapbox/appropriate-images-react';
```
katydecorah, line 1: Note: I'm not sure why this file is here, it appears unused, so let's get rid of it.

colleenmcginnis:
@katydecorah @danswick  here are some initial thoughts! @katydecorah happy to talk more on Thursday if it would be helpful. 

>Should we add filter by product? This PR doesn't include it yet as the options already feel robust.

I think the primary use of the filter will be different for “Examples” and “Tutorials”. For “Examples” I expect “topics” will be a common way for users to filter (also language for mobile examples). For tutorials, I expect users will want to filter by “products” so I think we should be prepared for the component to accommodate a pretty long list (maybe a “See all” option like below).

image


>Right now you can only select one item from each filter category.

With the current UI, I would expect to be able to select multiple items from each filter category, but I actually don’t know that selecting multiple items will be important to users (this would be a good thing to bring back to the docs team to discuss). If we decide to allow users to select multiple items, I expect it would significantly complicate the URL query string functionality (not sure how big of a con that is). If we decide not to allow users to select multiple items, I feel like we should consider a different UI that is more commonly used for selecting a single item (like a dropdown).

>Should we only provide filters for topics and videos? Is levels and languages too much?

I think it’s ok to have a lot of filter options for the “Tutorials” page because it does contain such diverse content. I also think the languages filter will be helpful on the mobile pages. In my opinion, we should plan to include them all to start and we can remove them later if they are overwhelming to users.

Also, are videos also included in the results when the toggle is off or will you only see videos if the toggle is on? If the former, we should make it more specific like "Videos only". 

>How should the pages be ordered by default?

For “Tutorials” I think it would be helpful to order by difficulty level (beginner to advanced). For “Examples” we should show examples with the topic “Getting started” first. After that I don’t think it matters.

katydecorah:
@colleenmcginnis Thanks for your responses, here are some follow-up.


>> Should we add filter by product? This PR doesn't include it yet as the options already feel robust.

> I think the primary use of the filter will be different for “Examples” and “Tutorials”. For “Examples” I expect “topics” will be a common way for users to filter (also language for mobile examples). For tutorials, I expect users will want to filter by “products” so I think we should be prepared for the component to accommodate a pretty long list (maybe a “See all” option like below).

Ok this makes sense.

From an implementation perspective, I think the easiest and most customizable option is to enable all filters if the data is available and then add frontMatter props to disable certain filters, for example `hideProductFilter`, as needed.

I’m wary about a “See all” option because it’s easy for the user to completely miss it.

> > Right now you can only select one item from each filter category.

> With the current UI, I would expect to be able to select multiple items from each filter category, but I actually don’t know that selecting multiple items will be important to users (this would be a good thing to bring back to the docs team to discuss). If we decide to allow users to select multiple items, I expect it would significantly complicate the URL query string functionality (not sure how big of a con that is). If we decide not to allow users to select multiple items, I feel like we should consider a different UI that is more commonly used for selecting a single item (like a dropdown).

I think that if we allow the user to select multiple items from each filter category, then we should consider not pushing values to the query string. I think this would be a good moment to decide: **what value does pushing the filters to the query string provide?** And if this is a feature we still want to consider.

Switching to dropdowns would provide a cleaner UI, but it is sometimes easier for the user to see all the options laid out — all though that sentiment can be negated when there are a lot of options. I can create mockup of what this could look like so we can discuss it at scrum today.

> Also, are videos also included in the results when the toggle is off or will you only see videos if the toggle is on? If the former, we should make it more specific like "Videos only".

I’ll update to “Videos only”,

> For “Tutorials” I think it would be helpful to order by difficulty level (beginner to advanced). For “Examples” we should show examples with the topic “Getting started” first. After that I don’t think it matters.

Would it work if we always filter by “Getting started” topic and then level? I could see this system becoming complicated if we build in different types of filtering.

---

It seems like we don't have an exact idea of what we want the filtering experience to look like. I think we should determine this at scrum today so that we have a plan moving forward.

katydecorah:
Per scrum today:

- We decided that the filter will only allow one option from each filter category to be selected at once. Based on the content for each filter category, we think it’s unlikely that a user would want to select more than one option from a given filter category. Even if they did, they’d be able to toggle the filter to see the different offerings.
- Since we’re only allowing one option per category, we’ll continue working on the URL query string feature to provide flexibility in the future.
- To keep this first iteration tightly scoped to a minimal viable product, we will use selects to display each filter that has options.

katydecorah:
@colleenmcginnis This PR is ready for review! 

- The cards look a little tight right now, but once #343 lands the content area will have more space.
- I made some adjustments to `src/components/page-layout/components/content.js` to create a `ContentWrapper` this allows us to easily move the filters into the PageLayout aside area. I found that the page looks much cleaner with the filters on the right-side of the page (for larger screens).
- I renamed `src/components/page-layout/components/layouts/example/content.js` to `src/components/page-layout/components/exampleIndex.js`. Once #343 lands, we won't have `layouts/components` and this new file name makes it easier to reference it when talking about these files/components.

src/helpers/batfish/filters.js
```diff
@@ -0,0 +1,110 @@
+// builds all the data you need to power exampleIndex layout
```
colleenmcginnis, line 1: I did not get to this file yet!

colleenmcginnis (commented):
@katydecorah @danswick this is looking good to me! I left some comments throughout. I ran out of time to take a close look at the filter Batfish helper, but I can take a closer look tomorrow!

docs/src/pages/guides/batfish-helpers.md
katydecorah, line 133:
```suggestion
  - If available, the object may have `videos: true` to indicate that some of its sub pages have `video: true` in the frontMatter.
```

src/helpers/batfish/navigation.js
```diff
@@ -168,5 +168,7 @@ function sortAlpha(arr) {
 
 module.exports = {
   buildNavigation,
-  findSection
+  findSection,
+  accordionSorter,
```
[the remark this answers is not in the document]
katydecorah: It was used by filters.js at one point, but no longer. I'll update this file.

src/components/page-layout/components/content.js
```diff
@@ -1,22 +1,47 @@
...
         frontMatter={frontMatter}
         sectionPath={parentPath}
         AppropriateImage={AppropriateImage}
       />
+    ) : (
+      <ContentWrapper {...this.props}>{children}</ContentWrapper>
     );
-  };
+  }
+}
```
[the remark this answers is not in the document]
katydecorah: Sure can!

src/components/page-layout/components/exampleIndex.js
```diff
@@ -0,0 +1,407 @@
...
+    cardColSize: PropTypes.number,
+    showFilters: PropTypes.arrayOf(filterOptions)
+  }).isRequired,
+  AppropriateImage: PropTypes.func
+};
+
+class FilterSection extends React.PureComponent {
+  render() {
+    const { title, data, activeItem, isSwitch, id, handleInput } = this.props;
+    const levelMap = {
```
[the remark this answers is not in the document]
katydecorah: Makes sense to me!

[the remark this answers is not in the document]
katydecorah: Actually! I just noticed that LevelIndicator component has a similar object. Let's import that object here instead :)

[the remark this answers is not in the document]
katydecorah: Here's the commit to make that happen: https://github.com/mapbox/dr-ui/pull/345/commits/335e4bf3764d9fa3b5f8aa82f2a4643972c49750

It required making a small adjustment to LevelIndicator to make the labels capitalized, so it added a few more files to the diff.

src/components/page-layout/components/exampleIndex.js
```diff
@@ -0,0 +1,407 @@
...
+      let searchParams = new URLSearchParams(window.location.search);
+      if (value) searchParams.set(label, value);
+      else searchParams.delete(label);
+      history.pushState(
+        null,
+        '',
+        window.location.pathname + '?' + searchParams.toString()
+      );
+    }
+  };
```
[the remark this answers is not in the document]
katydecorah: I'm not sure we've used the URLSearchParams API in docs before! Here are some links if it's helpful:

- https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams (This API says there is no IE support, but I tested in IE11 and it works 🤷. But since on line 86, we check that `URLSearchParams` exists in window, if a browser didn't support the API, it would skip the logic and not push the URL. We can consider this feature a progressive enhancement.)
- https://developer.mozilla.org/en-US/docs/Web/API/History/pushState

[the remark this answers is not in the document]
katydecorah: I added some more inline comments to help describe this function.

src/helpers/batfish/filters.js
```diff
@@ -0,0 +1,110 @@
...
+    (page.topic && page.topic === topic) ||
+    (page.topics && page.topics.indexOf(topic) > -1)
+  );
+}
+
+function sortBy(arr, prop) {
+  return arr.sort((a, b) => parseInt(a[prop]) - parseInt(b[prop]));
+}
+
+// sorts pages by "Getting started" topic(s), then by level, then alphabetically by title
```
colleenmcginnis, line 82: I think I suggested this order, but I'm worried we'll forget why later. I wonder if we should document here in a code comment and/or in the Batfish helpers guide that (even though the comment here is accurate) the _intent_ is to start with "Getting started" examples for product example pages **or** the beginner level tutorials on the tutorials page, then alphabetically by title.

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -116,22 +112,26 @@ module.exports = () => {
...
 
 ### Output
 
-- The shape of topics is an object, where the top-level keys are pathnames for top-level pages that have subpages with `topics` or `topic`.
-  - Each object has a `topics`. It contains a unique list of topics, ordered by count of pages with that topic.
-    - Each topic has `pages`. It contains metadata for each page that has that topic.
+- The shape of filters is an object, where the top-level keys are pathnames for `examplesIndex` pages.
+  - Each object has a `products` array, a unique list of products.
+  - Each object has a `topics` array, a unique list of topics.
+  - Each object has a `pages` array of all sub pages to display. The `pages` array is ordered by the `order` frontMatter property and then alphabetically by `title`.
```
colleenmcginnis, line 131: Is this still true given the `pageSorter` function in the filter helper?

colleenmcginnis (commented):
Just finishing up my earlier review! 💁‍♀️

docs/src/pages/guides/batfish-helpers.md
katydecorah: It's not, I'll update!

src/helpers/batfish/filters.js
katydecorah: Ok! I'll document!

colleenmcginnis (approved):
🍂

PR #346, opened 2020-11-02
[info arch] Rename ExamplesPage -> HelpPage

katydecorah:
This PR merges into #334 

This PR renames ExamplesPage component to HelpPage.

- We'll be adding the concept of "Filters" to examples page -- this code will live within PageLayout to starte.
- We will not apply filters to "Help" pages such as https://docs.mapbox.com/android/maps/help/ and instead will stick with the same layout (Tutorials and Troubleshooting sections with cards).
- This will allow us to render the `<HelpPage />` component on each `help.md` page and pass the dataset on the page instead of a JSON file that's tucked away to make it easier to make updates (alternatively we can still import the dataset, but this will help improve visibility on how the page is loaded).

There are no other physical changes, yet, only a rename to help set us up for Filter work.


colleenmcginnis:
I think you still need to rename `src/components/help-page/__tests__/__snapshots__/examples-page.test.js.snap` to `src/components/help-page/__tests__/__snapshots__/help-page.test.js.snap` to get tests to pass! The rest makes sense to me. 👍

katydecorah:
@colleenmcginnis 🤦 thanks! I updated the snapshot

PR #347, opened 2020-11-03
[info arch] Rework HelpPage to accept json to build cards

katydecorah:
**This PR builds on top of #345 and will merge into it once both PRs are reviewed.** Currently Help pages use the `examplesLayout` layout, but in #345 we are introducing filters, which the content on Help pages do not need. 

This PR reworks the HelpPage component to accept a `data` prop which is JSON data to build the card layout. In the previous iteration of `ExamplesPage` you had to supply the CardContainers and Cards, HelpPage will now build the CardContainers and Cards based on the data you supply.

This component is meant to make these help pages consistent:

- https://docs.mapbox.com/android/maps/help/
- https://docs.mapbox.com/ios/maps/help/
- https://docs.mapbox.com/unity/maps/help/
- https://docs.mapbox.com/mapbox-tiling-service/help/

During rollout, we'll update the above Help pages to look like `docs/src/pages/help/index.md` (We'll import HelpPage and import the help dataset, which should already be in the correct format to be passed as a `data` prop to the the component).

We'll also be able to adjust the data sent through to the "data" prop to only pass a certain dataset if we decide to split these "Help" pages into "Tutorials" and "Troubleshooting" pages. (Did we decide to do this during rollout?)

This PR also removes the `formatTopics` Batfish helper. This was introduced to help format data to work with examplesLayout, but this is no longer necessary. Instead, we can load the `HelpPage` component on a given page and pass the JSON data to the component.

## How to test

- View the HelpPage test cases app (`npm start`)
- View the catalog help page (`npm run start-docs` -> `http://localhost:8080/dr-ui/help/`)

NOTE: During review, please ignore the appearance of the left-side bar (or lack thereof). The updates to PageLayout in #344 will cover this.

## QA checklist


src/components/help-page/__tests__/help-page-test-cases.js
```diff
@@ -11,59 +10,78 @@ testCases.basic = {
...
-          />,
-          <Card
-            key="2"
-            title="Example two"
-            path="path"
-            description="Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
-          />
-        ]}
-      />
+    introText: '',
```
colleenmcginnis, line 43: This seems like a tricky pattern... If `introText` is not included as a prop, then the default text is used and to not include any `introText` you have to include an empty string. I wonder if we should just get rid of the default and hardcode it for each page? What do you think?

colleenmcginnis (approved):
Looks good! One thing to consider...

src/components/help-page/__tests__/help-page-test-cases.js
katydecorah: That makes sense, especially since we'll be splitting the Help pages into Tutorials and Troubleshooting -- I think this would have made sense if we were keeping Help pages in tact. I'll update to remove the prop all together.

PR #349, opened 2020-11-04
[info arch] Add OverviewHeader options to PageLayout

katydecorah:
This PR proposes a solution to the issue identified in https://github.com/mapbox/dr-ui/pull/344#issuecomment-721859813 where the OverviewHeader will fall below the OnThisPage section on smaller screens. If accepted, this PR will merge into #345.

This PR:
- Allows OverviewHeader props to be passed to PageLayout via frontmatter. If `frontmatter.overviewHeader` exists, then the page will render the OverviewHeader component above the OnThisPage component within PageLayout.
- Use flexbox on OverviewHeader to help improve sizing between content and image.
- Hide image sooner to prevent squishing content.
- Move tag under h1 (to the right of current version, changelog) since Tag has the same line height as this line of content.

## How to test

- View catalog page (`npm run start-docs` -> http://localhost:8080/dr-ui/components/)
- View OverviewHeader test case page (`npm test`)

## Screenshots

> image

> image

## QA checklist


docs/src/pages/components/index.md
```diff
@@ -10,24 +10,19 @@ hideTitle: true
...
-  - "const version = require('../../../../package.json').version; // eslint-disable-line"
+overviewHeader:
+  title: 'Dr. UI'
+  features:
+    - 'React components to build documentation sites'
+    - 'Support for IE 11 and all modern browsers'
+  changelogLink: /dr-ui/changelog/
+  installLink: https://github.com/mapbox/dr-ui/blob/main/README.md
+  ghLink: https://github.com/mapbox/dr-ui
+  # version: "" dynamic version is set in pages-shell.js
```
colleenmcginnis, line 22: We'll definitely want to be able to set this dynamically (using `constants.json`) including for multi-structured sites. That should be possible in the `PageShell` in each site, right? Just wanted to flag that in case that's a complexity you hadn't considered yet!

katydecorah: > That should be possible in the PageShell in each site, right?

Exactly! I tested this and it will work for multi-structure, too.

src/components/overview-header/overview-header.js
```diff
@@ -17,15 +17,18 @@ class OverviewHeader extends React.PureComponent {
...
+    const tag = props.tag && this.buildTag();
+
+    if (!versionEl && !changelogLinkEl && !tag) {
       return null;
     }
 
     return (
-      <p>
+      <div className="mb12">
+        {tag}
```
colleenmcginnis, line 28: I don't really like how the placement of the tag looks (I think it kind of gets lost and doesn't really align with the list elements), but I'm not sure if it's just because I'm used to it being next to the name of the product... 

Screen Shot 2020-11-05 at 3 01 39 PM

colleenmcginnis (approved):
I think this approach makes sense! I have one lil picky thing... but I could be convinced to let it go!

src/components/overview-header/overview-header.js
katydecorah: That's a fair point. I'll put it back to where it was!

PR #354, opened 2020-11-11
[Info arch] Allow appending tabs, including links to external pages, to nav tabs

danswick:
This PR adds an option for appending an arbitrary list of tabs to the nav accordion. This includes external links, to which `NavigationAccordion` will automatically add a "share" icon to denote that the link is to an external page.

image

## How to test

- Pull down this branch
- `npm run start-docs` to build the dr-ui docs site
- Observe the external links in the navigation accordion
- [optionally] try adding another page to the `addPages` variable in the docs site's `batfish.config.js` file

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 96% with Lighthouse version: `6.2.0`, specifically for batfish helpers guide.


src/components/navigation-accordion/navigation-accordion.js
```diff
@@ -65,6 +66,11 @@ export default class NavigationAccordion extends React.Component {
           style={{ letterSpacing: '0.025em' }}
         >
           {label}
+          {external && (
+            <span className="ml3">
```
katydecorah, line 70: Small styling suggestion: a slightly lighter color on the icon could help add distinction between the text and icon. And moving the icon 1px up will align it a little more with the text:

```suggestion
            <span className="ml3 color-darken50 relative" style={{top: '-1px'}}>
```

Current | Proposed
:---|:---
localhost_8080_dr-ui_components_ (3) | localhost_8080_dr-ui_components_ (1)

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -171,6 +172,12 @@ class PageShell extends React.Component {
 ]
 ```
 
+### Shape of appended pages
```
katydecorah, line 175: 🎸🎸🎸

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -25,6 +25,7 @@ These dataset functions often use the page's relative pathname as a unique ident
 - `siteBasePath`, required. The function requires the `siteBasePath`.
 - `data`, object. Provided by the data selector.
 - `sections`, array. See [Shape of multi-structured sections](#shape-of-multi-structured-sections).
+- `addPages`, array. See [Shape of appended pages](#shape-of-appended-pages).
```
katydecorah, line 28: Could you add this to "Shape of multi-structured `sections`" as well? Around line 150.

katydecorah, line 28: Maybe we should clarify that this addPages is for single structure only? Perhaps something like:

```suggestion
- `addPages`, array. See [Shape of appended pages](#shape-of-appended-pages). For single-structured sites only. For multi-structured sites, see [Shape of multi-structured sections](#shape-of-multi-structured-sections).
```

katydecorah (approved):
Lookin great! I left 2 documentation suggestions and 1 styling suggestion.

docs/src/pages/guides/batfish-helpers.md
danswick: Whew ok! I was definitely misunderstanding this implementation detail in a small but, it turns out, very significant way! 

I think it might be clearer for contributors if, instead of appending an `addPages` property to the `sections` object _before_ passing it to `buildNavigation`, we just support `addPages` in `buildMultiLevels`. So you would do something like: 

- `buildNavigation(basePath, data, sections, addPages)` for multi-structured sites.
- `buildNavigation(basePath, data, null, addPages)` for single-structured sites.

I made some changes here that show what I was thinking: https://github.com/mapbox/dr-ui/pull/354/commitshttps://github.com/mapbox/dr-ui/pull/354/files/b2c4c21714698ad9b55271f45d7fd917d056f6ee..28d34c76c32ee2e9342651782beda4b41f673434/57196a2fffba5c5cd493d3c242814fb255ab56c6.

danswick:
After looking at how `buildNavigation` is used for single- and multi-structured pages for a little while longer, I do think it will be worth moving to named parameters. When we roll out these changes, we'll need to tweak that function call anyway (adding `addPages` or `null`s as necessary), so it should only be a small bit of extra work to bracketize everything and we won't have any stray `null`s lying around!

src/helpers/batfish/navigation.js
```diff
@@ -93,10 +94,11 @@ function formatPages(siteBasePath, data, sections) {
     }));
 }
 
-function buildMultiLevels(sections, pages) {
+function buildMultiLevels(sections, pages, addPages) {
   return sections.reduce(
     (obj, section) => {
-      const sectionPages = pages.filter((f) => f.section === section.path);
+      let sectionPages = pages.filter((f) => f.section === section.path);
+      if (addPages) sectionPages = sectionPages.concat(addPages);
```
katydecorah, line 101:
```suggestion
      let sectionPages = pages.filter((f) => f.section === section.path);
      if (addPages) sectionPages = sectionPages.concat(section.addPages);
```

@danswick I think we still need to get `addPages` from the sections object, and not from the passed parameter.  I think our multi-structure test case is not catching this problem because we are only outputting one section (`maps`) and we should update it to use at least two sections.

For example, in iOS the site has many sections, let's just look at two:

```
[
      {
        path: 'maps',
        title: 'Maps SDK for iOS'
      },
      {
        path: 'navigation',
        title: 'Navigation SDK for Android'
      }
    ]
```

For each of these sections of the site, we want to add Tutorials and Troubleshooting links and each section will have its own product params appended to each link.

If we use `buildNavigation({siteBasePath, data, sections, addPages})` for iOS, how will `addPages` in this statement make sure each section adds the correct links to Tutorials and Troubleshooting? This code will append all `addPages` to every section.

This is why we created the `addPages` prop in `sections`  so that we can properly concat these new pages to the correct section.

katydecorah: If we want to use `addPages` like is committed, then I think we could add an optional prop to the addPages shape (`section`) to help identify which section the links belongs to. Example:

```js
const addPages = [
  {
    title: 'Tutorials',
    path: 'https://docs.mapbox.com/help/tutorials?product=maps+sdk+for+ios',
    navOrder: 4,
    section: 'maps'
  },
  {
    title: 'Troubleshooting',
    path: 'https://docs.mapbox.com/help/troubleshooting?product=maps+sdk+for+ios',
    navOrder: 5,
    section: 'maps'
  },
  {
    title: 'Tutorials',
    path: 'https://docs.mapbox.com/help/tutorials?product=navigation+sdk+for+ios',
    navOrder: 3,
    section: 'navigation'
  },
  {
    title: 'Troubleshooting',
    path: 'https://docs.mapbox.com/help/troubleshooting?product=navigation+sdk+for+ios',
    navOrder: 4,
    section: 'navigation'
  }
];
```

And then we could filter `addPages` and then concat them to `sectionPages`:

```js
const thisSectionPages = addPages.filter(f => f.section === section.path);
if (thisSectionPages) sectionPages = sectionPages.concat(thisSectionPages);
```

katydecorah (approved):
@danswick I tested locally on /android-docs and it works beautifully! I left one small suggestion about a test below for you to consider, but I think this is ready to merge!

PR #356, opened 2020-11-12
[info arch] Filter out pages in filter helper that have no title

katydecorah:
In attempting to test Mapbox GL JS, I received the error:

```
TypeError: Cannot read property 'localeCompare' of undefined
```

There are several redirect pages in Mapbox GL JS example folder and the Batfish filter function was attempting to include them in building the filters. Since redirects do not have title, it failed while attempting to sort.

This PR filters out all pages that do not have a title to prevent pages without the correct amount of context from being included in the filter.

## How to test

I created a reduced test case by carrying over three pages from Mapbox GL JS (the Examples index page, an examples redirect, and an examples page). I created a test to make sure that when filters are built with that dataset that no errors are thrown.


danswick (approved):
Looks good to me! Great catch.

PR #357, opened 2020-11-12
[Info arch] Add option to override <OnThisPage> default visibilty 

danswick:
This PR adds a new frontmatter property for layouts: `onThisPage`. If left unspecified, the default aside behavior will happen: aside will appear on pages with `layout: page`, but will not appear for any other layout. With this addition, `onThisPage: true` will force an aside to appear on pages with other layouts. `onThisPage: false` will cause asides to be hidden for pages with `layout: page`.

## How to test

In the catalog site, edit two pages: 

- For one of the guides with `layout: page`, add the following frontmatter: `onThisPage: false`. 
- For one of the examples, add the following frontmatter: `onThisPage: true`.
- Open the catalog site and verify that the first page you edited no longer contains an aside and that the second does!

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 98% with Lighthouse version: `6.2.0`.


src/components/page-layout/components/content.js
```diff
@@ -64,9 +64,18 @@ export class ContentWrapper extends React.Component {
...
     const headings = frontMatter.headings || this.props.headings;
-    const showToc = headings && headings.length > 0 && layout === 'page';
+    let showOnThisPage;
+    // Use `onThisPage` frontmatter to conditionally hide on page layouts or show elsewhere.
+    // By default, show `OnThisPage` only for 'page' layouts.
+    if (typeof onThisPage !== 'undefined') {
+      !onThisPage ? (showOnThisPage = false) : (showOnThisPage = true);
+    } else {
+      layout === 'page' ? (showOnThisPage = true) : (showOnThisPage = false);
+    }
```
katydecorah, line 76: @danswick I think we can streamline this logic by adding `onThisPage` as a layout option:

1. Add `onThisPage: true` to "page" layout object in `layout.config.js`. (This can replace `sidebar: toc` which is an old configuration.)
2. Then we can update the lines above with similar logic as the layout/frontMatter here to hide feedback: https://github.com/mapbox/dr-ui/blob/5c7c698c23644f75baa0f118bfb0063e9c2a0791/src/components/page-layout/components/content.js#L111-L114

Something like:

```js
// check frontmatter than default to layout config
const showOnThisPage = frontMatter.onThisPage ? frontMatter.onThisPage : layoutConfig.onThisPage;
```

katydecorah (approved):
💯

PR #359, opened 2020-11-13
[info arch] Create "icon" Tag variant; add Tag to NavigationAccordion

katydecorah:
During testing, we realized that the NavigationAccordion was missing the `Tag` prop. When we added Tag back to the component, we realized that the Tag did not fit well given that the NavigationAccordion in PageLayout was a small width. This PR adds a new `Tag` variant/prop `icon` that will display just an icon for the tag to conserve space.

- Add new `icon` prop to `Tag` that will display an icon instead of the Tag's text. `Themes` assigns all icons, but if using `custom` tag, you can use `customIcon` to assign whatever icon (from Assembly) you want.
- Add `Tag` to `NavigationAccordion` (this feature was in the previous version of this component and is being added back in). This `Tag` will use the `small` and `icon` Tag props to conserve space.
- Update `Note` image to use lightning icon to match `icon`. 

## Screenshots

> image
> Fundamentals tag in Navigation SDK for Android

> image
> Updated beta Note with "lighting" icon and all beta tag variants

@colleenmcginnis I know you suggested using the "lightning" icon for beta, want to confirm that we should also update the beta Note icon to match.

## How to test

Observe the following test cases app pages:
- /NavigationAccordion
- /Tag
- /Themes
- /Note - Using “lightning” for Beta icon

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `6.2.0`.

### DIFF
--- a/src/components/themes/__tests__/themes-test-cases.js
+++ b/src/components/themes/__tests__/themes-test-cases.js
@@ -70,7 +70,7 @@ const customProps = {
   customTooltipText: 'Contact us for access to this feature.',
   customStyles: {
     background: '#FEDADA',
-    color: '#DA2E30',
+    color: '#c12325',
     borderColor: '#FD8383'
   },
   customIcon: 'contact'
--- a/src/components/themes/__tests__/themes-test-cases.js
+++ b/src/components/themes/__tests__/themes-test-cases.js
@@ -25,8 +25,11 @@ testCases.beta = {
   description: 'beta',
   element: (
     <div>
-      <Note theme="beta">Beta note.</Note> <Tag theme="beta" />{' '}
+      <Note theme="beta">Beta note.</Note>
+      <Tag theme="beta" />
       <Tag theme="beta" small={true} />
+      <Tag theme="beta" icon={true} />
+      <Tag theme="beta" small={true} icon={true} />
     </div>
   )
 };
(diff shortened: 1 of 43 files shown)


src/components/themes/__tests__/themes-test-cases.js
```diff
@@ -63,9 +70,10 @@ const customProps = {
   customTooltipText: 'Contact us for access to this feature.',
   customStyles: {
     background: '#FEDADA',
-    color: '#DA2E30',
+    color: '#c12325',
```
katydecorah, line 73: This custom tag's contrast wasn't passing, so I updated it to make sure the Themes test cases app page has 100% on accessibility.

colleenmcginnis:
>@colleenmcginnis I know you suggested using the "lightning" icon for beta, want to confirm that we should also update the beta Note icon to match.

Let's do it!

danswick (approved):
This seems like a good change to me! Tested locally as well and all seems to be working as expected. I am a little concerned that the meaning of the lightning bolt icon won't be obvious to users, but I couldn't find anything that seemed to be widely used except the word "beta," so I'm cool with test driving the bolt!

katydecorah:
> I am a little concerned that the meaning of the lightning bolt icon won't be obvious to users

Yeah, I agree it's not really obvious. The concept of an icon for "beta" seems hard to come by. I think people often opt to just say "Beta" since it's short already. I'm my quick research, I saw an icon of a beaker (kinda like 🧪), but to me that reads more experimental. But yes, let's test drive for now and we'll rework as needed!

PR #360, opened 2020-11-13
[Info arch] Allow children to appear at the top of example index pages

danswick:
This PR allows contributors to place normal jsxtreme content in their examples index files. That content will be rendered above the examples content.

## How to test

Check out the examples index in this repo's docs site! 

image


katydecorah (approved):
💯

PR #362, opened 2020-11-17
[info arch] Improve filter and showFilters handling

katydecorah:
This PR fixes an issue where on some pages where `showFilters` is defined in the frontmatter and those filters do not have more than one option, the "Filters" heading appears:

> image

This PR cleans up the evaluation of each filter and then will remove an item from `showFilters` if it does not have more than one option. Also, fixes PropTypes `showFilter` as it was formatted incorrectly.

## How to test

Observe /PageLayout test case (`Example page layout, does not show filters because "showFilters" sets only "topics" and "topics" has one option.`)


danswick (approved):
💐 looks good to me!

PR #363, opened 2020-11-17
[info arch] Add "domain" prop to PageShell; remove old search-related props

katydecorah:
This PR adds a new `domain` prop to PageShell. The props accepts an object with `title` and `path` to allow you to customize the first item in the breadcrumb. This PR also removes unused search-related props.

## How to test

The "full layout" PageLayout test cases is now using the `domain` prop.


danswick (approved):
🍨 beaut!

PR #365, opened 2020-11-18
[Info arch] Adjust showfilters proptypes validation

danswick:
Fixes a small `propTypes` validation issue with `showFilters` frontmatter.

## How to test

Run the docs catalog site and check for console errors related to prop validation. Filters should also still work!

## QA checklist


katydecorah:
@danswick oops, I caught this in #362

PR #367, opened 2020-11-18
[info arch] Increase limiter--wide at 800px min-width query

katydecorah:
This PR overrides `.limiter--wide` width at the 800px min-width query to fill the page's width more. This will help give the page layout more space on smaller devices.

## How to test

Open the catalog page on a smaller monitor (1024 x 768). The `limiter-wide` class should have a width of `91.666666666%` (instead of `83.33333333%`)

## QA checklist

### DIFF
--- a/src/css/docs-prose.css
+++ b/src/css/docs-prose.css
@@ -4,6 +4,11 @@
     max-width: 1800px !important;
   }
 }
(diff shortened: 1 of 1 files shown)


danswick (approved):
💅

PR #368, opened 2020-11-18
[Info arch] Make Card description optional

danswick:
For info-arch rollout, we need to display some cards without descriptions. In order for prop validation to pass, the `description` prop needs to be optional.

### DIFF
--- a/src/components/card/__tests__/__snapshots__/card.test.js.snap
+++ b/src/components/card/__tests__/__snapshots__/card.test.js.snap
@@ -186,3 +186,98 @@ exports[`card Card with no image renders as expected 1`] = `
   </div>
 </a>
 `;
+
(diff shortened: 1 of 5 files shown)


katydecorah (commented):
@danswick What do you think about making these lines conditional (only display if `this.props.description` is available)? Otherwise each card would generate an extra div, not a big deal but on a site like Mapbox GL JS Examples that's about 116 extra divs 🤑 

https://github.com/mapbox/dr-ui/blob/7a32e88ecd3b4b8554990a875221ab958744b467/src/components/card/card.js#L47-L49

danswick:
Good idea, @katydecorah! I made that div conditional and added a test case.

katydecorah (approved):
@danswick Looks great! And thanks for cleaning up those props 💆‍♀️

PR #372, opened 2020-11-23
[info arch] Add unprose class to OverviewHeader ul

katydecorah:
This PR adds the `unprose` class to the OverviewHeader component's ul to allow us to better exclude this specific list from local list styles, Ref https://github.com/mapbox/studio-manual/pull/132


danswick (approved):
Seems like a smart change to me 👍

PR #375, opened 2020-11-23
[info arch] Add `flex-parent--center-cross` to OverviewHeader to center content

katydecorah:
This PR will vertically center the content in OverviewHeader to keep the left-side content (heading, features, buttons, etc) in alignment with the right-side (image).

## How to test

View the /OverviewHeader test case.


danswick (approved):
🦃

PR #379, opened 2020-11-30
[info arch] Fix path to files to work for test cases app and catalog page

katydecorah:
I noticed that some components on the catalog site have broken assets. This PR fixes the relative path to `files` so that the path will work for the test cases app and catalog page.

## How to test

- The catalog page shouldn't log to console images or videos as 404.
- The test cases app assets load as expected: /Browser, /CardContainer, /Card, /DemoIframe, /Phone, /RelatedPage, /Video


danswick (approved):
💅

PR #380, opened 2020-11-30
[info arch] Set height to prevent content shift as Search component loads

katydecorah:
This PR adds `h36` class to the div that wraps the Search component in PageLayout component. This will prevent the page from shifting while the search component loads.

## How to test

This may be tricky to test effectiveness locally as I've only been able to reproduce on staging. But I think checking the /PageLayout tests cases app page to make sure the Search component loads as expected is a good thing to check.

@danswick for review


danswick (approved):
Tested locally and this looks good to me! Once we get this onto a staging site we can check again just to be sure, but this seems pretty straightforward!

PR #381, opened 2020-12-09
Sort まず始めに (Getting started) to front of topic list

katydecorah:
This PR updates the filter Batfish helper to sort まず始めに (Getting started) to front of the topic list. 

## How to test

- Review the new test (with out the intervention `Webアプリ` would come before `まず始めに`)

@danswick for review

Ref https://github.com/mapbox/docs-jp/pull/55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Main
+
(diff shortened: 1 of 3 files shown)


danswick (approved):
Looks good to me! The test works during local testing 💅

PR #387, opened 2020-12-22
Update GLWrapper to check for IE

katydecorah:
This PR:

- Update mapbox-gl-supported to the latest.
- Reworks the message for GLWrapper to use `notSupportedReason` provided by mapbox-gl-supported:
  - If IE 11, show our message about IE 11 and Mapbox GL JS support.
  - If no WebGL support, show original GLWrapper message.
  - else let the user know the reason supplied by `notSupportedReason`. (There can be [many reasons](https://github.com/mapbox/mapbox-gl-supported/blob/214c1ed12dba316a6434469384010948354a8c69/index.js#L18-L30) why Mapbox GL is unsupported that are unrelated to WebGL or IE!)
- Adds an optional prop to GlWrapper: `reason`. This allows us to create (non-rendering) test cases for IE and no WebGL support to generate snapshot tests. This will be helpful in case future mapbox-gl-supported versions change the handling of these two reasons.

## How to test

Because Jest doesn't have WebWorker support, it makes it difficult to generate separate tests for IE 11 support, no WebGl support, and any other messages. (You'll noticed that the snapshot tests updated to state the reason for GL unsupported is due to `Mapbox GL is unsupported due to insufficient worker support.`)

You can also manually test the GlWrapper:

1. Run test-cases app and open /GlWrapper (npm start)
2. Open page in IE 11 via BrowserStack, it should look like:
image

One way to check for WebGL support is to disable it in Firefox (temporarily):

1. In Firefox go to `about:config`
2. Search for `webgl.disable` and set it to `true.
3. Open page in Firefox, it should look like:
image

## QA checklist

Fixes #387

### DIFF
--- a/src/components/gl-wrapper/__tests__/__snapshots__/gl-wrapper.test.js.snap
+++ b/src/components/gl-wrapper/__tests__/__snapshots__/gl-wrapper.test.js.snap
@@ -52,3 +52,155 @@ exports[`gl-wrapper Basic renders as expected 1`] = `
   </div>
 </div>
 `;
+
+exports[`gl-wrapper reason='insufficient ECMAScript 6 support' renders as expected 1`] = `
+<div
+  className="dr-ui--note py18 px18 round flex-parent flex-parent--row mb18"
+  style={
+    Object {
+      "background": "#feefe2",
+      "color": "#78471c",
+    }
+  }
+>
+  <div
+    className="flex-child mr18 none block-mm pt3"
+  >
+    <div
+      className="bg-orange-light round-full color-orange flex-parent flex-parent--center-main flex-parent--center-cross"
+      style={
+        Object {
+          "height": 50,
+          "width": 50,
+        }
+      }
+    >
+      <svg
+        className="events-none icon"
+        focusable="false"
+        role="presentation"
+        style={
+          Object {
+            "height": 40,
+            "width": 40,
+          }
+        }
+      >
+        <use
+          xlinkHref="#icon-alert"
+          xmlnsXlink="http://www.w3.org/1999/xlink"
+        />
+      </svg>
+    </div>
+  </div>
+  <div
+    className="flex-child prose"
+  >
+    <div
+      className="txt-bold txt-m mb6"
+    >
+      Mapbox GL unsupported
+    </div>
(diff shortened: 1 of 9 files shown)


src/components/gl-wrapper/__tests__/__snapshots__/gl-wrapper.test.js.snap
```diff
@@ -1,6 +1,142 @@
...
+  </div>
+  <div
+    className="flex-child prose"
+  >
+    <div
+      className="txt-bold txt-m mb6"
+    >
+      Mapbox GL unsupported
+    </div>
+    Mapbox GL is unsupported due to insufficient worker support.
```
katydecorah, line 51: 📝 We expect the "basic" test case app example to return this message as Jest isn't a browser and doesn't have WebWorker support.

src/components/gl-wrapper/__tests__/__snapshots__/gl-wrapper.test.js.snap
```diff
@@ -1,6 +1,142 @@
...
+  </div>
+  <div
+    className="flex-child prose"
+  >
+    <div
+      className="txt-bold txt-m mb6"
+    >
+      Mapbox GL unsupported
+    </div>
+    Starting with v2.0.0, Mapbox GL JS is no longer compatible with any version of Internet Explorer. If you need to support Internet Explorer, consider using the
```
katydecorah, line 104: 📝 This non-rendering test case example sets the reason manually to `insufficient ECMAScript 6 support` so we can track the generated warning message for IE.

src/components/gl-wrapper/__tests__/gl-wrapper-test-cases.js
```diff
@@ -9,4 +10,22 @@ testCases.basic = {
...
+
+noRenderCases.webgl = {
+  description: "reason='insufficient WebGL support'",
+  element: (
+    <GLWrapper reason="insufficient WebGL support">
+      Your browser supports Mapbox GL!
+    </GLWrapper>
+  )
+};
+
```
katydecorah, line 30: 📝 This `noRenderCases` generate a snapshot so we can track if there are any changes the next time we update mapbox-gl-supported.

colleenmcginnis (approved):
🎉

PR #392, opened 2021-01-11
Add Lazy component

### LINKED ISSUE #388: Create `Lazy` component
I think we should add a new component: `Lazy`.

This component would be used to lazy load another component. If we decide to start lazy loading Mapbox GL maps, we could include this new component in GLWrapper to handle lazy loading maps, if supported by the browser. I can imagine other components that load a significant amount of data or other packages using this new component to help improve performance.

The component will include:
- The ErrorBoundary component as a fallback if the dynamic import fails
- A loader to display until the component loads

Resources:
- https://reactjs.org/docs/code-splitting.html#reactlazy

katydecorah:
_This PR merges in to the `perf` branch #393 ._

This PR adds a new component, `Lazy`. It's a wrapper for https://loadable-components.com/ which is the React team's recommended alternative for lazy loading components server side (`React.lazy` is not supported on the server side by React, yet). 

`Lazy` accepts two props:

- `lazyComponent` a function to import the desired component
- `lazyHeight` the height of the imported component to reduce content layout shift during load.
- `lazyClasses` is an alternative to `lazyHeight` in case the height of the component is variable, you can define classes to best describe the height of the component.

You must define `lazyHeight` or `lazyClasses`.

You can also pass props to `Lazy` that will be passed along to the import component. The lazy props are named distinctly to avoid prop name collisions.

The component will display a loader while the `lazyComponent` loads. 

Fixes #388 

## How to test

- [ ] Open test cases app (`npm start`) at /Lazy. The example is lazy loading the Search component. Confirm that the Search component loads as expect.

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `6.4.0`.

### DIFF
--- a/docs/src/pages/guides/lazy.md
+++ b/docs/src/pages/guides/lazy.md
@@ -1,0 +1,13 @@
+---
+title: Lazy loading
+description: Learn when to use lazy loading.
+order: 5
+layout: page
+contentType: guide
+products:
+  - Documentation
+prependJs:
+  - "import Lazy from '../../../../src/components/lazy/lazy';"
+---
+
+{{<Lazy lazyClasses="h30-mm h36" lazyComponent={() => import('../../../../src/components/search/examples/basic.js')} />}}
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "@mapbox/rehype-prism": "^0.5.0",
     "@mapbox/remark-config-docs": "^0.9.0",
     "@mapbox/remark-lint-link-text": "^0.5.0",
-    "@mapbox/remark-lint-mapbox": "^2.2.0",
+    "@mapbox/remark-lint-mapbox": "^2.4.0-perf.0",
     "babel-eslint": "^10.1.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "babelify": "^10.0.0",
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "@elastic/react-search-ui": "1.3.2",
     "@elastic/react-search-ui-views": "1.3.2",
     "@elastic/search-ui-site-search-connector": "1.3.2",
+    "@loadable/component": "^5.14.1",
     "@mapbox/mapbox-gl-supported": "^2.0.0",
     "@mapbox/mr-ui": "^0.9.1",
     "@sentry/browser": "^5.27.4",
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -101,6 +101,7 @@
     "@elastic/react-search-ui": "1.3.2",
     "@elastic/react-search-ui-views": "1.3.2",
     "@elastic/search-ui-site-search-connector": "1.3.2",
+    "@loadable/component": "^5.14.1",
```
katydecorah, line 104: ✏️  Documentation: https://loadable-components.com/

Supports: v5.x (React v16.3+): IE11, IE 9+ (with Map + Set polyfills), all evergreen browsers

colleenmcginnis (commented):
@katydecorah is there a good way to slow down loading to see the fallback in action? Or will that be easier to see when we test this out on one of our sites?

katydecorah:
>  is there a good way to slow down loading to see the fallback in action? Or will that be easier to see when we test this out on one of our sites?

@colleenmcginnis it's a little challenging on the test cases app! I just pushed a temporary demo page to the catalog site. Here's a good way to catch the loader: 

1. Pull down this branch
2. Run `npm run build-docs && npm run serve-static` 
3. Open http://localhost:8080/dr-ui/guides/lazy/
4. Open Network tab with the following settings and then hit refresh
image

docs/src/pages/guides/lazy.md
```diff
@@ -0,0 +1,13 @@
...
+order: 5
+layout: page
+contentType: guide
+products:
+  - Documentation
+prependJs:
+  - "import Lazy from '../../../../src/components/lazy/lazy';"
+---
+
+{{<Lazy lazyClasses="h30-mm h36" lazyComponent={() => import('../../../../src/components/search/examples/basic.js')} />}}
```
katydecorah, line 13: ✏️ Added this as a testing demo. This page will be the future home of lazy loading documentation -- to be completed in a future PR.

katydecorah:
@colleenmcginnis I just pushed a beta release of a fix for the xtreme test (https://github.com/mapbox/remark-lint-mapbox/pull/59) this should fix the tests! It's good we caught this by making this demo as we'd very likely see this error in all other sites.

colleenmcginnis (approved):
This looks great to me! I also had fun reading about lazy loading. 🤓 Thanks for the links!

PR #393, opened 2021-01-12
⚡️ Performance main branch

katydecorah:
This PR is the main branch for performance and accessibility updates to dr-ui. 

## To do

- [x] Add required height to MapWrapper to improve CLS (cumulative layout shift).
- [x] Remove IE 11 support including poyfills.

Fixes https://github.com/mapbox/dr-ui/issues/389
Fixes https://github.com/mapbox/dr-ui/issues/391

PR #395, opened 2021-01-13
Set minHeight in MapWrapper (formerly GLWrapper) and DemoIframe

katydecorah:
_This PR merges into #393._

This PR:
- Renames GLWrapper to MapWrapper to better describe how you can use the component.
- Requires a `height` for MapWrapper to prevent a layout shift while mapbox-supported makes a decision.
- Required a `title` for DemoIframe to improve accessibility.

## How to test

* [ ] Open /MapWrapper in test cases app
* [ ] Open /DemoIframe in test cases app
* [ ] Preview these maps/pages on staging (I created a beta release from this branch and pushed to /studio-manual https://github.com/mapbox/studio-manual/pull/138):
  * [ ] /studio-manual/examples/3d-buildings/  
  * [ ] /studio-manual/guides/

## QA checklist

### DIFF
--- a/src/components/demo-iframe/demo-iframe.js
+++ b/src/components/demo-iframe/demo-iframe.js
@@ -4,10 +4,6 @@ import ChevronousText from '@mapbox/mr-ui/chevronous-text';
 import MapWrapper from '../map-wrapper/map-wrapper';
 
 export default class DemoIframe extends React.Component {
-  static defaultProps = {
-    gl: true
-  };
-
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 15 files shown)


katydecorah:
This is ready for review @colleenmcginnis, I updated the original post with some reviewing suggestions.

src/components/demo-iframe/demo-iframe.js
```diff
@@ -1,60 +1,24 @@
...
+      />
     );
-
-    return props.gl ? <GLWrapper>{contents}</GLWrapper> : contents;
   }
 }
 
 DemoIframe.propTypes = {
-  src: PropTypes.string.isRequired, // absolute URL src for iframe
-  gl: PropTypes.bool, // set to false if the iframe does not use GL
```
colleenmcginnis, line 58: Does removing the `gl` prop prevent us from using this component with mapbox.js maps? Or we just don't care if an error message is shown for mapbox.js maps because we won't be supporting browsers that don't work with Mapbox GL JS anyway?

katydecorah: Ah, I didn't think of that use case! Maybe we should revert to adding the `gl` prop back.

katydecorah:
@colleenmcginnis this is ready for review again. Per chat, we will be pausing the lazy load work for when we're able to dedicate more time to updating Batfish to support this feature. In the meantime, I've reworked the goal of this PR:

1. Improve CLS (cumulative layout shift) for maps by requiring `height` prop.
2. Improve accessibility of iframes by requiring `title` prop.

I updated the original post again (reviewing section is pretty much the same).

You can see the CLS improvement most significantly on /studio-manual/examples/3d-buildings/. Moving one needle at a time!

Before | After
--|--
image | image

src/components/demo-iframe/demo-iframe.js
```diff
@@ -26,35 +22,53 @@ export default class DemoIframe extends React.Component {
...
     // if the src has an `access_token` value, replace it with mapboxAccessToken
     // otherwise use the src as is
-    const src = makeRequest
-      ? props.src.replace('MapboxAccessToken', this.state.mapboxAccessToken)
-      : props.src;
+    src = makeRequest
+      ? src.replace('MapboxAccessToken', mapboxAccessToken)
+      : src;
+
+    const iframeHeight = 400;
```
colleenmcginnis, line 40: Should this be a default prop value instead? Otherwise defining the height like in the [last test case](https://github.com/mapbox/dr-ui/pull/395/files#diff-3ccba7f92bbe8bb3c98f3082aa91935efade014657d4e9860f50dd8127ce7dcdR38 ) won't work, right? Or I guess the alternative would be to remove `height` from the test case and always use a height of 400. I don't have a strong opinion either way!

katydecorah: @colleenmcginnis Oops, didn't mean to add it to that testcase! I'll remove. We had DemoIframes to always be 400px and I haven't heard a reason to change that. Unless you think we should make it variable?

colleenmcginnis: Nope! Let's leave it at 400px. I do love consistency ... including in the heights of my iframes!

colleenmcginnis (approved):
🖼️

PR #398, opened 2021-01-19
Require AppropriateImages in `vimeoThumbnail` prop in `RelatedPage`

katydecorah:
*This PR merges into #393* 

This PR makes `vimeoThumbnail` require AppropriateImages rather than a string. This will help improve performance of these thumbnail images by making sure they are compressed.

## How to test

- [x] View the /RelatedPage test case (`npm start`) and make sure image loads
- [x] View /accounts/guides/settings/#setting-up-saml-sso-for-your-mapbox-account on staging and make sure image loads

Note:
- I needed to add video-thumb generated images to the the catalog page and test case files to get the path to work correctly in path environments. A little annoying, but I couldn't find a better workaround!

## QA checklist


katydecorah:
@colleenmcginnis This will likely be the last dr-ui PR  for this quest (unless something comes up during rollout)!

katydecorah:
@colleenmcginnis whoops, there may be _one_ more if we decide to drop IE 11 support (to be determined in sync today)!

colleenmcginnis (approved):
Makes sense to me! It's kind of annoying that we'll need to import the repo-specific `AppropriateImage` component in every md file that uses `RelatedPage` to link to a video, but I can't think of any way to get around that...

katydecorah:
> It's kind of annoying that we'll need to import the repo-specific AppropriateImage component in every md file that uses RelatedPage to link to a video

Yeah, I agree! I think during this work with compression we'll start to find more things that aren't great about rendering images and other assets and I will start to write them down. It's possible that we are out growing the AppropriateImage system, but it's the best method we have right now to make sure images are optimized. I think once we have full compliance, it'll be easier to either find a new system or make updates to the current one!

Issue #415, opened 2021-03-04
Proposal: add `sitemapIgnore` Batfish helper


katydecorah:
In https://github.com/mapbox/batfish/pull/348, we are adding an `ignoreFile` sitemap option. With this option, we'll be able to exclude generated asset files from the sitemap on sites like Mapbox GL JS and MTS. We recently learned that these assets files were added (incorrectly) to the sitemap and resulted in Google Search console errors. 

For implementation we have two options:

1. Update affected sites with latest Batfish, create `sitemap-ignore.js` file to exclude the assets folder, and update `batfish.config.js` to set the `ignoreFile` option. 
See the following commit for sample implementation: https://github.com/mapbox/mts-docs/pull/90/commits/94d55d858cb14caf9dba4af426ef3b8a37177dca

2. Create a `sitemapIgnore` Batfish helper that will generate the `sitemap-ignore.js` file on build to include any files with `hideFromSearchEngines: true` in the frontmatter, in addition to custom fixes such as excluding the assets folder. 
See the following commit for sample implementation: https://github.com/mapbox/mts-docs/pull/90/commits/46528a76db22c59284f6251d87279040042bafab

My preference is Option 2. I think if we're are going to generate a sitemap, then we should show it good hygiene.

Option 2 will solve the original problem flagged by Google Search console in addition to removing all paths we do not want to include in the sitemap. Option 2 would fix all the errors in the screenshot below, while Option 1 would only fix `Submitted URL not found (404)`:

![image](https://user-images.githubusercontent.com/2180540/110021370-a19bad80-7cf8-11eb-8ac1-a9462accd711.png)

----

This fix isn't urgent and rollout to all sites could happen during the next site-wide rollout, but we could bump MTS and Mapbox GL JS early to fix the 404 errors.

@mapbox/docs let me know if you have any feedback on this proposal!

PR #427, opened 2021-04-05
Create `prepareSitemap` Batfish helper to improve sitemap

katydecorah:
**This branch will merge into #431** 

This PR adds a `prepareSitemap` Batfish helper to improve sitemap output. The function will automatically remove pages with `hideFromSearchEngines: true` and `splitPage: true` from the sitemap. This will prevent URLs we don't want users to visit from appearing in the sitemap.

Fixes https://github.com/mapbox/dr-ui/issues/415

## How to test

There are tests in this PR that assert the output of this function and the catalog's site's sitemap.

## QA checklist

Not applicable for this function.

## Before merge


.travis.yml
```diff
@@ -6,3 +6,5 @@ notifications:
 scripts:
   - npm test
   - npm run build-docs
+  # run sitemap tests after build to assert sitemap is correct
+  - jest src/helpers/batfish/__tests__/sitemap.test.js
```
danswick, line 10: Nice!

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -252,6 +253,69 @@ class PageShell extends React.Component {
 {{JSON.stringify(splitPages, null,2)}}
 ```
 
+## Prepare sitemap
+
+`prepareSitemap` creates a data file that lists all files with `hideFromSearchEngines: true` or `splitPage: true` set in the page's frontMatter. Use this generated file with Batfish's [`sitemap.ignoreFile`](https://github.com/mapbox/batfish/blob/main/docs/configuration.md#sitemap) option to generate a sitemap that excludes any paths in the prepared file.
```
danswick, line 258: Do you only need to worry about `prepareSitemap` if you're planning on hiding some files from search engines? Or will all dr-ui / Batfish sites need the `sitemapIgnore: ... prepareSitemap()` combo?

docs/src/pages/guides/batfish-helpers.md
```diff
@@ -252,6 +253,69 @@ class PageShell extends React.Component {
...
+### Set up in batfish.config.js
+
+```js
+const { prepareSitemap } = require('@mapbox/dr-ui/helpers/batfish/index.js');
+
+module.exports = () => {
+  return {
+    sitemap: {
+      // generated by the sitemapIgnore dataSelector
+      ignoreFile: '_site_tmp/data/sitemap-ignore.js'
```
danswick, line 276: Similar question here: you can use `sitemap: true` if you don't want to exclude some pages from the sitemap, right?

danswick (approved):
This looks like a great helper, @katydecorah. I just had a couple of questions about how implementation would work for sites that don't need to exclude any pages.

docs/src/pages/guides/batfish-helpers.md
katydecorah: You only need to introduce this Batfish helper if you have pages that you want to exclude, otherwise the sitemap will build without intervention.

docs/src/pages/guides/batfish-helpers.md
katydecorah: If you don't want to intervene with the sitemap, you don't need to specific `sitemap` at all. For Batfish, the sitemap will automatically build if you define `siteOrigin`, ref: https://github.com/mapbox/batfish/blob/main/docs/configuration.md#sitemap

PR #430, opened 2021-04-06
Load Search with a facade

katydecorah:
**This branch will merge into #431** 

This PR replaces the initial entry point to the Search component with a facade. The facade is a button that looks exactly like the Search component, but only in looks. Once the users clicks or focuses on the facade, the full Search component will load via dynamic import. 

A few things were moved around to support the facade:
1. The logic to load the Swiftype moved to a new file: `search-provider.js`. When the user interacts with the facade, we dynamically load the `search-provider.js`.
2. The search input and search button are now components that are shared by `search-box.js` and `search-facade.js`. This allows us to have a seamless transition.

## How to test

Overall: you should not be able to tell that the Search component loads a facade until you click or focus on it.

1. `npm start` and visit the /Search test cases app.
    - [ ] On small screens, test cases the use the search button switch to input
    - [ ] Each test case returns a search result
2. `npm run start-docs`
    - [ ] Search component works in PageLayout

## QA checklist

- [x] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `7.0.0`.

## Before merge


src/components/search/search.js
```diff
@@ -1,101 +1,72 @@
...
+      loadSearch: false,
+      useModal: !this.props.disableModal || !this.props.resultsOnly
+    };
+  }
+
+  loadSearch = () => {
+    this.setState({ loadSearch: true });
+  };
+
+  /* Use SearchInput on smaller screens, SearchButton on larger screens (unless disableModal) */
```
katydecorah, line 23: Moved this logic out from search-box.js

katydecorah:
Performance testing in docs-lighthouse looks good. I want to wait for #414 to merge so that I can test these together.

src/components/search/search-box.js
```diff
@@ -149,47 +131,13 @@ class SearchBox extends React.PureComponent {
...
-                    ref={(input) => {
-                      this.docsSeachInput = input;
-                    }}
-                    placeholder={this.props.placeholder}
-                    className={classnames('input bg-white', {
-                      'px60 h60 txt-l': this.state.useModal,
-                      'px36 h36': !this.state.useModal
-                    })}
-                    {...getInputProps({
+                  <SearchInput
```
katydecorah, line 134: 💡 Created SearchInput and SearchButton components (at bottom of this file) so the SearchFacade could use them, too.

src/components/search/search-provider.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { SearchProvider, WithSearch } from '@elastic/react-search-ui';
+import { SearchBox } from './search-box';
+
+class Search extends React.PureComponent {
```
katydecorah, line 6: 💡 The contents of this file was split from search.js

src/components/search/__tests__/search-test-cases.js
```diff
@@ -42,7 +43,7 @@ testCases.narrow = {
   element: <Narrow />
 };
 
-testCases.withConnector = {
+noRenderCases.withConnector = {
```
katydecorah, line 46: 💡 I switched this to a noRender test cases because it uses a dummy Swiftype site connector and will not return results

src/components/search/search.js
```diff
@@ -1,101 +1,94 @@
...
+    super(props);
+    this.state = {
+      SearchProvider: undefined,
+      useModal: !this.props.disableModal || !this.props.resultsOnly
+    };
+  }
+
+  /* Wait to load the full Search component */
+  loadSearch = () => {
+    import(
```
katydecorah, line 19: 💡 It can be difficult to know when the SearchProvider loads (which is good 😎), while testing you could add a console.log statement in this function to verify when the facade gets swapped out.

danswick (approved):
This is awesome, @katydecorah! I walked through each of the test cases, tried your console.log suggestion, and ran the test site as well. Thanks to your helpful inline comments, I don't really have any questions or comments. This seems really well thought-out and solid to me 👍

PR #439, opened 2021-04-16
[Experiment] Enable Lighthouse CI to check test cases (accessibility, SEO, and best practices)

katydecorah:
This PR tests out adding a new Travis check that would run Lighthouse on the test cases app. It would only check accessibility, SEO, and best practices audits since many performance audits would not be effective on the test cases app.

## Notes

Initially, this PR was the run all Lighthouse audits (except PWA) on the catalog site, but since the performance audit varies it would require 3-5 run on each page. Targeting the testcases app will help us improve our components in other ways.

## To do

- [x] What's the average Travis build time before these tests versus after? Do we feel comfortable with adding time to each test build?
- [x] Can we run Lighthouse only on components that have changes in the diff? This could help reduce the Travis build time.

Status | Travis build time 
:---|---:
Before Lighthouse | ~3 mins
After Lighthouse (all test cases) | ~10 mins
After Lighthouse (1 test case) | ~4 mins
After Lighthouse (4 test cases) | ~4.5 mins

- [x] Fix failing audits


.travis.yml
```diff
@@ -1,10 +1,18 @@
 language: node_js
-sudo: false
```
katydecorah, line 2: 💡 this is deprecated

.travis.yml
```diff
@@ -1,10 +1,18 @@
...
 notifications:
   email: false
-
-scripts:
```
katydecorah, line 6: 💡 `script` is the preferred property name

Issue #245, opened 2020-02-07
Analytics quest


katydecorah:
It would be great to find opportunities to add events to interactions found in dr-ui components. For example, making sure every copy button in a code snippet sends a "Copied example" event to Segment. I think this data is really useful in understanding which features users interact with the most (or least).

PR #442, opened 2021-04-19
Enable react/jsx-no-bind

katydecorah:
**This branch merges into #431** 

This PR sets `react/jsx-no-bind: error` to improve performance. As a result of this eslint rule, this PR also:

- Sets `onCopy` on code snippet components to prevent requiring contributors to send an empty arrow function to get the CopyButton to appear. This will deprecate `onCopy` for `NumberedCodeSnippet` component. From a repo-scan, all sites were using `onCopy={() => {}}` with `NumberedCodeSnippet`. This will also add Segment track events to every code snippet so we can collect usage analytics.
- Creates breaking change for `highlighter` prop to remove the arrow function.

Ref https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

## How to test

Verify that the interactive pieces of the following components work in the test cases app:

- [x] AndroidLayoutCodeBlock
- [x] BackToTopButton
- [x] CodeSnippet
- [x] CodeToggle
- [x] ContextlessAndroidActivityToggle
- [x] ContextlessIosActivityToggle
- [x] Edit
- [x] Feedback
- [x] Highlight
- [x] NavigationAccordion
- [x] NumberedCodeSnippet
- [x] PageLayout (example filters)
- [x] RelatedPage (video icon)
- [x] Search
- [x] ToggleableCodeBlock

## Before merge


src/components/code-snippet/on-copy.js
```diff
@@ -0,0 +1,6 @@
+export default function onCopy() {
+  console.log('Copied example with clipboard');
```
katydecorah, line 2: 💡 This console.log is for testing, we must remove it before merging.

```suggestion
```

src/components/numbered-code-snippet/numbered-code-snippet.js
```diff
@@ -249,7 +254,6 @@ export default class NumberedCodeSnippet extends React.PureComponent {
     const highlightElements = [];
     const copyElements = [];
     let previousCount = 0;
-    let liveChunkCount = -1; // Incremented to give CopyButtons an identifier
```
katydecorah, line 252: :bulb: Since all instances of NumberedCodeSnippet where using an empty arrow function of onCopy, the chunk identifier hasn't been in use. I also don't think we need it, I think we'd rather know that users are copying code and not necessarily which code block.

src/components/code-snippet/on-copy.js
```diff
@@ -0,0 +1,6 @@
+export default function onCopy() {
```
katydecorah, line 1: 💡 This PR introduces a shared onCopy function to be used by all code snippet components. I think we could eventually do more with shared event tracks, but will save that for https://github.com/mapbox/dr-ui/issues/245

src/components/numbered-code-snippet/numbered-code-snippet.js
colleenmcginnis: I agree!

colleenmcginnis (approved):
I tested all impacted components in the test app and reviewed the diff -- nothing surprising in either!

PR #444, opened 2021-04-23
Fix prepareSitemap function to append index.html to files ending in index.js

katydecorah:
While testing #431, I noticed that in the prepareSitemap Batfish helper that all files that end in `index.js` must have `index.html` appended to the path (not just the root index.js file). This PR makes that fixes and adds a test case to support it.

Before this fix, the test case would return the following paths to ignore:

```json
["/_site/index.html", "/_site/core/", "/_site/java/", "/_site/core/guides/hidden/"]
```

This would result in the sitemap excluding everything under `/_site/core/` and `/_site/java/` since it uses globs. This is not what we want. We only want the `index.html` to be removed from the sitemap, not all pages under those directories.

With the fix, the paths are more specific:

```json
["/_site/index.html", "/_site/core/index.html", "/_site/java/index.html", "/_site/core/guides/hidden/"]
```

### DIFF
--- a/src/helpers/batfish/__tests__/sitemap.test.js
+++ b/src/helpers/batfish/__tests__/sitemap.test.js
@@ -64,6 +64,55 @@ describe('prepareSitemap', () => {
     ]);
   });
 
+  it('append index.html to files ending in index.js', () => {
+    const ignoreFile = prepareSitemap({
+      pages: [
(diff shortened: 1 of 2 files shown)


danswick (approved):
Great catch, @katydecorah!

PR #448, opened 2021-05-06
Fix props on search toggle

katydecorah:
This PR fixes an error in the Search component when clicking the site toggle.

cc @colleenmcginnis


katydecorah:
@colleenmcginnis Thanks for catching this! I pushed this beta release to /account staging and the Search toggle is working as expected.

colleenmcginnis (approved):
🦔 🔍

PR #462, opened 2021-06-21
Redesign Feedback

katydecorah:
This PR redesigns the Feedback component.

## QA checklist

## Before merge


colleenmcginnis (commented):
I took a close look at the different categories and just have one question below. I still need to spend some more time with:
- [ ] `src/components/feedback/feedback.js`
- [ ] `src/components/feedback/segment.js`
- [ ] `src/components/feedback/sentry.js`

src/components/feedback/categories.js
```diff
@@ -0,0 +1,81 @@
...
+import CategoryConfusing from './components/category-confusing.js';
+
+// This function holds the majority of text for the Feedback component
+// It returns an object:
+// - The key is the category's name
+// - `helpful` infers if the user found the content helpful
+// - `component` returns the component for the category
+// To add, remove, or change the categories, update the object and the Feedback will automatically update with your changes.
+export function categories({ type, submitFeedback }) {
+  const genericType = returnGenericType(type);
```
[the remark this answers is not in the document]
katydecorah: This will be variable, you can see this snapshot to see how the content will appear for different content types: https://github.com/mapbox/dr-ui/pull/462/files#diff-10c599eb89dae61a0ebabeefaab0ac598765b263a076a5ea08d0c9893507c3d7

> If yes, will it be difficult to report on trends in mode reports?

This is a good thing to call out, we could:

1. Adjust the value we send to Segment/Sentry to "Easy to understand" while keeping the label variable `The ${genericType} is easy to understand`.
2. Change the label to `It is easy to understand` to remove the variable content type.

What do you think?

HeyStenson (approved):
Things are working as expected! A few UI items I noted:

- A user can submit what is essentially blank feedback using the spacebar. Not sure if there's any way around this (or if it's even worth trying to circumvent behavior that likely won't be that common), but wanted to flag it just in case.
- The link to contact support opens in the same tab, rather than in a new tab. I know that there are some accessibility issues around opening in a new tab, but I worry that opening in the same tab may interrupt the user's documentation experience. I don't have super strong feelings about this, though, so if the pros outweigh the cons I'm fine with it as is!
- There doesn't seem to be an event fired when a user clicks out of the menu using the _x_ rather than continuing with the flow. I suspect that's on purpose since we wouldn't really get any additional info from a 'close' action, but wanted to flag just in case. 

The new feedback module looks AMAZING and I can't wait to see it in action! Thank you so much for all your hard work on this!

src/components/feedback/categories.js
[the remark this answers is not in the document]
colleenmcginnis: I like the simplicity/consistency of 2, but I think the specificity of 1 is helpful to guide the users to provide clear feedback. I lean slightly toward 1, but could easily be convinced otherwise if you feel differently!

src/components/feedback/__tests__/shared.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Asserts that the textarea will show an error message when the character limit is met
+// and the submit button becomes disabled
+export function textTooLong(feedback) {
+  const textarea = feedback.find('textarea');
+  // Add too much text
+  textarea.simulate('change', {
+    target: {
+      value: `${"I found a sandwich and I want to know why there isn't any mayonnaise.".repeat(
```
colleenmcginnis, line 16: 🥪

src/components/feedback/feedback.js
```diff
@@ -1,321 +1,222 @@
...
+          <div className="align-center prose relative">
+            <div className="mt-neg30 inline-block">
+              <BookImage size={75} />
+            </div>
+            <p>
+              <strong>Thank you!</strong>
+            </p>
+            <p>
+              Our documentation team will read your feedback. Thank you for
+              helping us improve this page.
```
colleenmcginnis, line 207: Should this `page` be a variable, too? Or is that overly complicated for something that won't add any clarity to the feedback submitting process?

src/components/feedback/segment.js
```diff
@@ -0,0 +1,58 @@
...
+    contactSupport
+  } = state;
+  const { site, section, location } = props;
+  // set user if available else set anonymousId (needed for forward-event request)
+  const identity = {
+    ...(user && user.id ? { userId: user.id } : { anonymousId: anonymousId })
+  };
+  return {
+    event: 'Sent docs feedback',
+    ...identity,
```
colleenmcginnis, line 30: Do we need the `username` and `anonymousId` both here and in line 51?

src/components/feedback/sentry.js
```diff
@@ -0,0 +1,52 @@
...
+          data: { plan: user.plan.id }
+        })
+    });
+  }
+  // configure data to send with feeedback
+  Sentry.configureScope((scope) => {
+    // set tag for site name
+    scope.setTag('site', site);
+    // set tag for category
+    scope.setTag('category', category);
```
colleenmcginnis, line 34: Not sure why the `category` and `categoryType` do not match in this [test feedback in Sentry](https://sentry.io/organizations/mapboxcom/issues/2498147667/?environment=staging&project=1848287&query=is%3Aunresolved). `category` is `Something is confusing` while `categoryType` is all the `I like this..` options. 

image

I was clicking a lot of things when I was testing, but there are only three events with this `sessionId` and the only `category` that appears is `Something is confusing`.

src/components/feedback/categories.js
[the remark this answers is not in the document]
katydecorah: I'll scope out 1!

src/components/feedback/feedback.js
katydecorah: Ohhh I think we can variabl-ize it!

katydecorah:
> A user can submit what is essentially blank feedback using the spacebar. Not sure if there's any way around this (or if it's even worth trying to circumvent behavior that likely won't be that common), but wanted to flag it just in case.

@HeyStenson Good catch! I know in the current Feedback the user must submit at least 3 characters for the submit button to appear. We can add that logic back in.

> The link to contact support opens in the same tab, rather than in a new tab. I know that there are some accessibility issues around opening in a new tab, but I worry that opening in the same tab may interrupt the user's documentation experience. I don't have super strong feelings about this, though, so if the pros outweigh the cons I'm fine with it as is!

From an accessibility standpoint, the same tab is preferred. But it is going to a different site altogether 🤔 @mapbox/docs what do you think?

> There doesn't seem to be an event fired when a user clicks out of the menu using the x rather than continuing with the flow. I suspect that's on purpose since we wouldn't really get any additional info from a 'close' action, but wanted to flag just in case.

Right, we aren't capture when the user closes out. I think we'll assume that they exited early, but that may be difficult to query. I wonder if we should add a `status` column to track their progress in the flow. 🤔 I'll consider this more!

src/components/feedback/sentry.js
katydecorah: I'll look into this!

src/components/feedback/categories.js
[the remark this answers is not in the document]
katydecorah: Option 1  was a fairly easy thing to implement 😅  c075e16!

src/components/feedback/__tests__/shared.js
katydecorah: 🏞️

src/components/feedback/segment.js
katydecorah: We do! Fixing!

katydecorah: Oh! But we only need 1 (userId or anonymousid) for this particular line.

src/components/feedback/sentry.js
katydecorah: I wonder if you tested "I like this" with the first /Feedback test case and then (without refreshing) tested the other test case with "Something is confusing" - I was able to replicate doing that. I think we can close / clear Sentry after submission. I'll see if that'll work. This will be important to get right for pages with sectioned feedback.

katydecorah: OH! https://docs.sentry.io/platforms/javascript/enriching-events/scopes/#local-scopes gotta use local scopes!!!

katydecorah: This should be fixed now! If you agree: smash that resolve conversation button.

katydecorah:
> There doesn't seem to be an event fired when a user clicks out of the menu using the x rather than continuing with the flow. I suspect that's on purpose since we wouldn't really get any additional info from a 'close' action, but wanted to flag just in case.

> > Right, we aren't capture when the user closes out. I think we'll assume that they exited early, but that may be difficult to query. I wonder if we should add a status column to track their progress in the flow. 🤔 I'll consider this more!

@HeyStenson I thought about this more and I think we should send an event to Segment when they close! I think that'll be interesting data and make it way easier to query exits.

katydecorah:
>> A user can submit what is essentially blank feedback using the spacebar. Not sure if there's any way around this (or if it's even worth trying to circumvent behavior that likely won't be that common), but wanted to flag it just in case.

> I know in the current Feedback the user must submit at least 3 characters for the submit button to appear. We can add that logic back in.


Since positive feedback is not required for some checkboxes, then I think the validation will need to happen when they click submit. Example: 

1. user inputs a space in textarea and clicks submit.
2. the textarea throws a validation error requesting they say more.

src/components/feedback/components/forms.js
```diff
@@ -0,0 +1,97 @@
...
+          <ControlTextarea
+            themeControlTextarea="textarea hmin120 bg-white"
+            themeLabel="txt-m mb6"
+            id={id}
+            label={label}
+            value={value}
+            onChange={this.handleFeedback}
+            placeholder={placeholder}
+            validationError={
+              (validationErrorMinimum ? 'Tell us more' : '') ||
```
katydecorah, line 48: @HeyStenson Do you have any suggestions for the error message that will appear when the user's feedback does not meet the minimum? I'm not sure if "Tell us more" is helpful here or needs a little more instruction 🤔

src/components/feedback/components/forms.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ControlTextarea from '@mapbox/mr-ui/control-textarea';
+import classnames from 'classnames';
+
+// character limit for the feedback textarea
+export const feedbackLimit = 1000;
+export const feedbackMinimum = 3;
```
katydecorah, line 8: @HeyStenson @colleenmcginnis What do you think about this minimum for required textarea? I carried this number over from https://github.com/mapbox/dr-ui/blob/920eb460aa7911eb49f692253b5104e3c5b25f60/src/components/feedback/feedback.js#L243. I think increasing it could add friction, but could help enforce more words from the user.

katydecorah:
## To do

- [ ] add Sentry fingerprint to each feedback to prevent merging, this will be important for positive feedback that doesn't have required text.
- [ ] find out if we can auto-assign "I like" feedback to a Sentry team (in this case #positive-feedback).

colleenmcginnis:
>find out if we can auto-assign "I like" feedback to a Sentry team (in this case #positive-feedback).

BRILLIANT!

katydecorah:
@colleenmcginnis I think this is ready for your review again!

src/components/feedback/components/forms.js
HeyStenson: Hmmm, been thinking about this for a lil while and I really can't think of anything better 😅  The only punchup I can think of that might make a difference is "Tell us more!" to make it seem more upbeat. (Other option would be "We need more detail" but I don't think that strikes the right tone.)

src/components/feedback/components/forms.js
HeyStenson: The shortest meaningful word I can think of is "No" but if someone added that as text input there's no scenario in which that would actually be useful information ... so I think 3 characters is good! If we find that we're getting poor quality feedback, we could up the value ⬆️

src/components/feedback/feedback.js
```diff
@@ -1,321 +1,230 @@
...
+          <div className="align-center prose relative">
+            <div className="mt-neg30 inline-block">
+              <BookImage size={75} />
+            </div>
+            <p>
+              <strong>Thank you!</strong>
+            </p>
+            <p>
+              Our documentation team will read your feedback. Thank you for
+              helping us improve this {this.props.type}.
```
colleenmcginnis, line 215: Do you need something like this since `type` is not a required prop? (Also `this.prop.type` is already set to `type` in line 167 right?)

```suggestion
              helping us improve this {type ? type : "content"}.
```

src/components/feedback/sentry.js
colleenmcginnis: SMASH.

colleenmcginnis (approved):
One small question. Everything else looks great to me, @katydecorah! Nice work. 👑

src/components/feedback/feedback.js
katydecorah: While `type` is not a require prop the default is "page" so I don't think we need this logic unless we want to be more specific about the text here.

Here are examples of how the current code will look:

- sectioned feedback: Thank you for helping us improve this section on AccessTokens.
- examples: Thank you for helping us improve this example.
- default: Thank you for helping us improve this page.

colleenmcginnis: Sounds good!

src/components/feedback/components/forms.js
katydecorah: I like that plan! Leaving as is! Thanks!

katydecorah:
Today I will create a beta release and roll out to a few repositories to check for bugs.

katydecorah:
here we go!!!

PR #469, opened 2021-07-28
Add Sentry tracing with new `sentryPerformance` prop on AnalyticsShell

katydecorah:
This PR will automatically add Sentry's tracing integration to the AnalyticsShell Sentry initialization via a new prop `sentryPerformance`. You can disable tracking with an empty object `<AnalyticsShell sentryPerformance={{}} />`.

When we are ready to enable Sentry performance trial, all sites will be ready to launch. 

## How to test

This will be difficult to test since we don't have the trial yet, but since tracing is automatic I think it should be ok!

https://docs.sentry.io/platforms/javascript/performance/

## QA checklist

## Before merge

### DIFF
--- a/src/components/analytics-shell/analytics-shell.js
+++ b/src/components/analytics-shell/analytics-shell.js
@@ -2,6 +2,7 @@ import React from 'react';
 import { Helmet } from 'react-helmet';
 import PropTypes from 'prop-types';
 import * as Sentry from '@sentry/browser';
+import { Integrations as TracingIntegrations } from '@sentry/tracing';
 import env from './env';
 
 export default class AnalyticsShell extends React.PureComponent {
@@ -11,6 +12,7 @@ export default class AnalyticsShell extends React.PureComponent {
       disableSentry,
       webAnalytics,
       sentry,
+      sentryPerformance,
       mbxMetadata
     } = this.props;
     // set mapbox metadata before initializing analytics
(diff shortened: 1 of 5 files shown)


katydecorah:
I deployed this alpha version to studio-manual staging.

src/components/analytics-shell/analytics-shell.js
```diff
@@ -59,6 +62,12 @@ AnalyticsShell.defaultProps = {
   sentry: {
     dsn: 'https://6ba8cfeeedad4fb7acb8576f0fd6e266@sentry.io/1384508'
   },
+  sentryPerformance: {
+    // This enables automatic instrumentation (highly recommended)
+    integrations: [new TracingIntegrations.BrowserTracing()],
+    // To set a uniform sample rate
+    tracesSampleRate: 0.2
```
danswick, line 69: What do you think about making the sampling rate configurable in case we need to tweak for various sites without doing a dr-ui release? The [Sentry docs](https://docs.sentry.io/platforms/javascript/performance/#verify) recommend testing with a rate of 1 so we may at least want to cover that use case. What do you think?

danswick (approved):
Left one suggestion for you, but otherwise this seems solid to me!

src/components/analytics-shell/analytics-shell.js
katydecorah: Per chat, these value are configurable. This array represents the default prop.

PR #473, opened 2021-08-12
Fix error handling of `forwardEvent` callback

katydecorah:
This PR fixes an issue where the `forwardEvent` callback was sending an event to Sentry regardless of error. By checking `if (err)` we can confirm that the callback is an error and then send it to Sentry. 

@danswick for review

Fixes https://github.com/mapbox/dr-ui/issues/472

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.1.1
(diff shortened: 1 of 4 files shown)


danswick (approved):
Looks great! 🌭

Issue #338, opened 2020-09-24
Create DownloadButton component


katydecorah:
We should consider adding a DownloadButton component as we use the following markup often, especially in /help:

```js
{{
<Button href="/help/demos/store-locator/marker.png" passthroughProps={{ download: "marker" }} >
    <Icon name='arrow-down' inline={true} /> Download custom marker
</Button>
}}
```

katydecorah:
Another use case for this component:

https://github.com/mapbox/atlas-docs/pull/539#issuecomment-720519128

cc @HeyStenson

katydecorah:
We did this in #474!

PR #474, opened 2021-08-18
Add DownloadButton component 

malwoodsantoro:
Addresses https://github.com/mapbox/dr-ui/issues/338

## QA checklist

## Before merge


src/components/download-button/download-button.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import Button from '@mapbox/mr-ui/button';
+import IconText from '@mapbox/mr-ui/icon-text';
+import PropTypes from 'prop-types';
+
+// eslint-disable-next-line react/prefer-stateless-function
+export default class DownloadButton extends React.Component {
```
katydecorah, line 7:
```suggestion
export default class DownloadButton extends React.PureComponent {
```

malwoodsantoro:
Additional examples that use download button: 

ZIP - https://docs.mapbox.com/data/traffic/guides/example/
image

ZIP - https://docs.mapbox.com/ios/maps/guides/install/#add-the-dependency(iOS + Android) 
image

malwoodsantoro:
@katydecorah would love to get your initial thoughts on my approach here. I tried passing className: 'unprose' in passThroughProps within the Button component, but it removed all of the styling from the button. I couldn't figure out quite how to target _just_ the button text without also removing the button styling. I ended up adding the CSS from https://github.com/mapbox/help/blob/publisher-production/src/css/site.css#L21-L29 to docs-prose.

src/components/download-button/download-button.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import Button from '@mapbox/mr-ui/button';
+import IconText from '@mapbox/mr-ui/icon-text';
+import PropTypes from 'prop-types';
+
+export default class DownloadButton extends React.PureComponent {
+  render() {
+    const { fileType, href } = this.props;
+    const fileName = href
```
katydecorah, line 9: Would you mind walking me through what's happening on lines 9-13?

src/css/docs-prose.css
```diff
@@ -127,6 +127,16 @@
...
+a.btn.btn--blue {
+  color: #fff;
+  text-decoration: none;
+  margin: 18px 0;
+}
+
+a.btn.btn--blue:hover {
+  color: #fff;
+}
+
```
katydecorah, line 139: This CSS will effect all elements with `btn btn--blue` classes. I think we'll want to be more specific here to prevent unintended side effects to other elements.

We have a couple options:

1. Prepend `.prose` to line 130 and 136. However this could effect other buttons.
2. Use passthroughProps in Button: `passthroughProps={{ download: fileName, className:'unprose btn btn--blue round-full py12 px24 txt-s my18' }}`. While we would need to rebuild the button classes to pass `unprose` to the button, we don't need to maintain CSS nor be concerned if the classes will effect other elements.

My preference is towards option 2, I like that we can maintain the styling for the component in one place and not rely CSS in docs-prose.css. What do you think?

src/components/download-button/__tests__/download-button-test-cases.js
```diff
@@ -0,0 +1,53 @@
...
+    href: '../files/test-zip.gz'
+  }
+};
+
+testCases.prose = {
+  description: 'DownloadButton with prose class',
+  props: {
+    fileType: 'PNG',
+    href: '../files/shop-15.png'
+  },
```
katydecorah, line 45: Since we are using `element` and are passing these props to the element in line 48, we no longer need this `props` object.

src/components/download-button/download-button.js
```diff
@@ -0,0 +1,34 @@
...
+        <Button href={href}>
+          <IconText iconBefore="arrow-down">Download {fileType} file</IconText>
+        </Button>
+      );
+    } else {
+      return (
+        <Button href={href} passthroughProps={{ download: fileName }}>
+          <IconText iconBefore="arrow-down">Download {fileType} file</IconText>
+        </Button>
+      );
```
katydecorah, line 26: There is repeated code here that I think we can optimize (repeated code often leads to human error and more maintenance). 

I see the difference between the two is that one has `passthroughProps`. 
- What might this look like if we added the React prop (`passthroughProps`) conditionally?
- Or, is there a downside to always setting `passthroughProps`? I don't actually know this, but if all file formats work with `passthroughProps`, then this would save us from having to add/maintain logic.

src/components/download-button/download-button.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+}
+
+DownloadButton.propTypes = {
+  href: PropTypes.string.isRequired,
+  fileType: PropTypes.oneOf(['CSV', 'PDF', 'PNG', 'ZIP']).isRequired
```
katydecorah, line 33: I love an opinionated PropType!

I am wondering if we should include... more? 

- JPG
- GeoJSON
- JSON
- GIF
- HTML
- CSS
- JS


**Or, maybe the most future friendly approach is to allow any value here (`PropTypes.string.isRequired`)?** 

By using `PropTypes.string.isRequired`, we don't have to:
- maintain a list of possible file extensions
- make updates to this component
- ignore PropTypes warnings in repositories that use a `fileType` outside this list

While we lose the ability to enforce consistency, it will make the component more flexible.

**OR! What if we didn't include a `fileType` prop and extract the file type from the `href`?**

While this will add a small bit of code to the component, we:
- reduce the need for one more propType for the contributor to have to remember to add
- can restyle (letter up/down-casing) the file extension to match our style guide as needed
- make the component future friendly


What do you think?

katydecorah (commented):
This is a really excellent start @malwoodsantoro and you are headed in the right direction. I left some suggestions of things to consider. Let me know if you want to discuss them more!

src/components/download-button/download-button.js
katydecorah: Ooof!

I glossed over the "Download for iOS" button (ref )https://github.com/mapbox/dr-ui/pull/474#issuecomment-902425956). I love using this button for that use case, but it will change a few of the ideas above. 

To support that use case we could have these propTypes:

- `href` (required)
- `text` (optional) 
  - If entered, use this value as the button's text. Example `text="Download for iOS"` results in `Download for iOS`. 
  - If empty, the default button text will be `Download {extension} file`. Example `Download PNG`. We could extract the extension from the `href` prop.

src/components/download-button/download-button.js
malwoodsantoro: @katydecorah this is helpful as it's making me really break this down a bit! I was searching for solutions to pull _just_ the filename without the extension for file types that require `passThroughProps`. I found this Stack Overflow article that was helpful - https://stackoverflow.com/questions/43638163/get-filename-from-url-and-strip-file-extension/43638276. 

This code takes the path and separates it at `/` with split(). It then uses pop() to remove the last element from the array which should be the filename. The replace() function is then used with a regular expression to find the beginning of the extension (`.`) and remove it.

src/components/download-button/download-button.js
malwoodsantoro: >What might this look like if we added the React prop (passthroughProps) conditionally?

Good call! I reworked this to keep the code from repeating.
 
> Or, is there a downside to always setting passthroughProps? I don't actually know this, but if all file formats work with passthroughProps, then this would save us from having to add/maintain logic.

If I use `passthroughProps` with a CSV, the file extension is removed from the downloaded file. 

CSV with `passthroughProps` 👎 
image

CSV without `passthroughProps` 👍 
image

I can definitely do some more testing here!

src/components/download-button/download-button.js
malwoodsantoro: _I just realized I can remove `toString` though since `href` is already a string_

src/components/download-button/download-button.js
malwoodsantoro: > OR! What if we didn't include a fileType prop and extract the file type from the href?

Love this so much! For _"can restyle (letter up/down-casing) the file extension to match our style guide as needed"_, can you point me to the best place to find this information in the style guide? I do see ZIP referenced in the Word list -> https://hey.mapbox.com/documentation/style-guide/word-list/ 🕺 

I added an optional text prop to support the iOS use case and future download buttons that require custom text 👍

src/css/docs-prose.css
malwoodsantoro: I am very much for option two as well so we don't have to worry about any unintended consequences in the future ✨  

> I like that we can maintain the styling for the component in one place and not rely CSS in docs-prose.css

💯

src/components/download-button/download-button.js
malwoodsantoro: Good news! I think we can simplify this a bit more and remove the need for the file type conditional as you suggested. Using `passthroughProps` with the filename (including extension) for _all_ file types seems to be the best option here. 

If I remove `passthroughProps` entirely, the PDF and PNG will open in a new browser window rather than downloading directly. I think the above CSV issue was happening because the browser wasn't detecting and appending the file type with just the filename specified in the `download` attribute (just for CSVs it seems). When using `passthroughProps` with the filename and extension, the extension is added as expected. This works for all test cases!

I guess the only thing to look into further would be if there are any downsides to `passthroughProps` 🕵️

src/components/download-button/download-button.js
katydecorah: > For "can restyle (letter up/down-casing) the file extension to match our style guide as needed", can you point me to the best place to find this information in the style guide? I do see ZIP referenced in the Word list -> https://hey.mapbox.com/documentation/style-guide/word-list/


That's exactly where I would point you to. We can also check with Heather and Dom!

src/components/download-button/download-button.js
```diff
@@ -0,0 +1,31 @@
...
+          {text || 'Download ' + fileType + ' file'}
+        </IconText>
+      </Button>
+    );
+  }
+}
+
+DownloadButton.propTypes = {
+  href: PropTypes.string.isRequired,
+  text: PropTypes.string
```
katydecorah, line 30: One last thing we'll want to do is document these props! 

We do this by adding a comment above each prop. The comment must being with double stars. Example:

```suggestion
/** Link to the asset that will be dowloaded. */
  href: PropTypes.string.isRequired,
  text: PropTypes.string
```


During the build process the catalog site will automatically add these comments to the property table. Example:
https://mapbox.github.io/dr-ui/components/#tag

https://github.com/mapbox/dr-ui/blob/c0df94f7677261e84f2173af361247bc0055918d/src/components/tag/tag.js#L55

2021-08-23 at 10 53 AM

src/components/download-button/download-button.js
```diff
@@ -0,0 +1,31 @@
...
+    return (
+      <Button
+        href={href}
+        passthroughProps={{
+          download: fileName,
+          className: 'unprose btn btn--blue round-full py12 px24 txt-s my18'
+        }}
+      >
+        <IconText iconBefore="arrow-down">
+          {text || 'Download ' + fileType + ' file'}
```
katydecorah, line 21: What do you think about dropping the word "file"? We can also ask the team to weigh in!

```suggestion
          {text || 'Download ' + fileType}
```

src/components/download-button/download-button.js
malwoodsantoro: SO COOL

src/components/download-button/__tests__/__snapshots__/download-button.test.js.snap
```diff
@@ -0,0 +1,85 @@
...
+    >
+      Download CSV
+    </span>
+  </span>
+</a>
+`;
+
+exports[`DownloadButton DownloadButton for PNG renders as expected 1`] = `null`;
+
+exports[`DownloadButton DownloadButton with optional text renders as expected 1`] = `null`;
```
katydecorah, line 43: These are rendering as null, I don't think we expect that. This likely means that we need to check-in on how we're generating the snapshot in download-button.test.js.

src/components/download-button/__tests__/download-button.test.js
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  describe(testCases.png.description, () => {
+    let testCase;
+    let wrapper;
+    let tree;
+
+    beforeEach(() => {
+      testCase = testCases.png;
+      wrapper = renderer.create(testCase.element);
```
katydecorah, line 28: The `png` and `zip` testcases do not have an `element` property so that's why the snapshot tests are returning null. Check out this test case for how we pull in test cases that use `component` and `props` instead of `element`:

https://github.com/mapbox/dr-ui/blob/c0df94f7677261e84f2173af361247bc0055918d/src/components/book-image/__tests__/book-image.test.js#L22-L38

docs/src/categories.json
```diff
@@ -28,6 +28,7 @@
     "PageLayout",
     "OnThisPage",
     "BackToTopButton",
+    "DownloadButton",
```
katydecorah, line 31: Nit: I wonder if DownloadButton would be better categorized as a "Features" rather than "Layout" - the Layout components typically are found within the PageLayout and/or have a more structural utility to them. What do you think?

katydecorah (commented):
We're so close!

PR #481, opened 2021-09-03
Upgrade to Assembly v1 🌝

malwoodsantoro:
- [x] Update repository to Node 14 (update package.json and .nvmrc) and npm 7 (update README.md requirements section)
- [x] Update assembly, mr-ui
- [x] Check docs-prose.css for styles that can be removed or adjusted:
  - [x] Headings: v1 uses docs font-size and line-heights. We can also consolidate the double selectors to`#docs-content .prose h*`.
  - [x] `sticky` is now in v1
  - [x] Color overrides, use /dr-ui/guides/a11y/ page to check color contrast
  - [x] `limiter--wide` limiter is deprecated in v1. Do we need this style or can we move it inline? (This is specific to PageLayout, but maybe should be addressed ahead?)
  - [x] txt-spacing0` is now in v1
  - [x] `scroll-auto-m*` in v1 scroll-auto is now `overflow-auto`. We should update these styles/classes to match.

## QA checklist

- [ ] No errors logged to console.
- [ ] Accessibility score in [Chrome DevTools Audit/Lighthouse](https://developers.google.com/web/tools/lighthouse#devtools) is 100% with Lighthouse version: `#.#.#`.
- [ ] Component is accessible at mobile-size viewport.
- [ ] Component is accessible at tablet-size viewport.
- [ ] Component is accessible at laptop-size viewport.
- [ ] Component is accessible at big-monitor-size viewport.

Open the test cases app locally on:

- [ ] Chrome, no errors logged to console.
- [ ] Firefox, no errors logged to console.
- [ ] Safari, no errors logged to console.
- [ ] Edge, no errors logged to console.
- [ ] Mobile Safari, no errors logged to console.
- [ ] Android Chrome, no errors logged to console.

## Before merge

- [ ] Add entry to CHANGELOG.md to describe changes.
- [ ] If updating dependencies or creating a release, run `npx browserslist@latest --update-db` to update the [browser data](https://github.com/browserslist/browserslist#browsers-data-updating) and commit any changes to package-lock.json.

PR #491, opened 2021-09-13
[Assembly v1] CodeSnippetTitle

malwoodsantoro:
This PR merge into https://github.com/mapbox/dr-ui/pull/481 and updates the CodeSnippetTitle component to Assembly v1.

- [x] Run codemod and look out for additional changes
- [x] Check docs-prose.css for any component-specific classes that should be updated or removed, including changes to the color scale.
- [x] Visually check test cases app for component

## QA checklist

## Before merge

### DIFF
--- a/src/components/code-snippet-title/code-snippet-title.js
+++ b/src/components/code-snippet-title/code-snippet-title.js
@@ -28,12 +28,12 @@ export default class CodeSnippetTitle extends React.PureComponent {
   render() {
     const { link, toggle } = this.props;
     return (
-      <div className="flex-parent-mm flex-parent--space-between-main-mm mb6">
-        <div className="flex-child">
+      <div className="flex-mm flex--space-between-main-mm mb6">
+        <div>
           {this.renderFilename()}
           {link && this.renderLink()}
         </div>
-        {toggle && <div className="flex-child">{toggle}</div>}
+        {toggle && <div>{toggle}</div>}
       </div>
     );
   }
--- a/src/components/code-snippet-title/code-snippet-title.js
+++ b/src/components/code-snippet-title/code-snippet-title.js
@@ -5,15 +5,13 @@ import Icon from '@mapbox/mr-ui/icon';
 export default class CodeSnippetTitle extends React.PureComponent {
   renderFilename = () => {
     return (
-      <div className="inline-block txt-bold mb6 color-gray-dark">
-        {this.props.filename}
-      </div>
+      <div className="txt-bold color-gray-dar mb6">{this.props.filename}</div>
     );
   };
 
   renderLink = () => {
     return (
-      <div className="inline-block ml12-mm">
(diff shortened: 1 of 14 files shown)


malwoodsantoro:
@katydecorah I'm running into some issues with snapshots updates included here that aren't a result of the changes in my PR. I think there may be an issue with my workflow: 

1) Checkout `assembly-v1-update` 
2) `git pull`
3) `git checkout -b assembly-v1-code-snippet-title assembly-v1-update`
4) Make updates to component file 
5) Run `npx jest -u`

Should I be pointing only to my component directory when running `npx jest -u`?

katydecorah:
@malwoodsantoro I'm sorry these steps aren't giving you the right result! Let's dig into:

> 4. Make updates to component file

Can you share what your process is for this step? The snapshot tests are failing because it looks like the codemod had run on other components outside the scope of the changes to `CodeSnippetTitle`. Is it possible that you ran codemod for the entire site, but only committed changes to CodeSnippetTitle? That could explain why when you run `npx jest -u` it generated snapshots with unrelated changes to CodeSnippetTitle.

When you run codemod, try (Replace `/Your/Path/To/` with your local path):

```
npx jscodeshift -t /Your/Path/To/mapbox/studio/codemods/assembly.js src/components/code-snippet-title/
```

This will make codemod only run on the code-snippet-title component.

Then you can run:

```
npx jest -u
```

Let me know how it goes! We can also pair today!

src/components/code-snippet-title/code-snippet-title.js
```diff
@@ -28,12 +28,12 @@ export default class CodeSnippetTitle extends React.PureComponent {
   render() {
     const { link, toggle } = this.props;
     return (
-      <div className="flex-parent-mm flex-parent--space-between-main-mm mb6">
-        <div className="flex-child">
+      <div className="flex-mm flex--space-between-main-mm mb6">
+        <div>
```
katydecorah, line 32: I noticed an issue that I think it present in the current implementation: on smaller screens we want the file name and GitHub link to stack.

I think we could make that happen by adding `flex-mm` here and then removing `inline-block` from line 8 (may want to remove `mb6` from line 8 as well) and `inline-block` from line 16.

```suggestion
        <div className="flex-mm mb6">
```

Or we could update `ml12-mm` on line 16 to `ml12` to prevent the loss of space between the filename and GitHub link on smaller screens.

katydecorah (approved):
🌼 (Left one comment to consider!)

src/components/code-snippet-title/code-snippet-title.js
malwoodsantoro: image

@katydecorah thank you for the suggestions. Let me know what you think! I kept `mb6` since it looks better when the filename and link are stacked to have some space in between them.

Issue #457, opened 2021-06-07
Update babel.config.js to enable ES modules


katydecorah:
We should set `[['@babel/preset-env', { modules: false }]` in babel.config.js to take advantage of Webpack's tree shaking:

> When you use ES modules, webpack becomes able to do tree-shaking. Tree-shaking is when a bundler traverses the whole dependency tree, checks what dependencies are used, and removes unused ones.

> Warning: Don’t accidentally compile ES modules into CommonJS ones.
> 
> If you use Babel with babel-preset-env or babel-preset-es2015, check the settings of these presets. By default, they transpile ES’ import and export to CommonJS’ require and module.exports. Pass the { modules: false } option to disable this.

Size | Before | After
:---|:---:|---:
package |  77.0 kB  | 74.1 kB                                                   
unpacked |  341.2 kB  | 290.3 kB 

https://developers.google.com/web/fundamentals/performance/webpack/decrease-frontend-size#use_es_modules

PR #498, opened 2021-09-15
Do not preserve es modules in test environment

katydecorah:
We recently [enabled es modules](https://babeljs.io/docs/en/babel-preset-env#modules) in #457, but https://github.com/mapbox/react-test-kitchen (the system for our test cases) works best with common js. This PR updates the `test` babel environment to not preserve es modules.

I noticed this when trying to interact with the /Search test case and the dynamic import failed to load.

## How to test

1. Run: `nvm use && npm ci && npm start`
2. Visit: /Search test cases
3. Interact with search, there should be no errors logged to the console (re: the dynamic import should load as expected).

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -7,8 +7,9 @@ module.exports = {
     'babel-plugin-transform-react-remove-prop-types'
   ],
   env: {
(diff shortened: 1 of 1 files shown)


malwoodsantoro (approved):
Looks great 👍 no errors!

PR #524, opened 2021-09-29
[Assembly v1] Replace viewport and scroll classes in PageLayout

katydecorah:
While reviewing /mapbox-gl-js/style-spec/layers/#symbol I noticed that the right navigation bar did not have a scrollbar. (For reference https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#symbol does!) This is because I missed updating the deprecated viewport and scroll classes 😄


malwoodsantoro (approved):
🍏

PR #542, opened 2022-12-20
improve search

chriswhong:
Improves the UX and styling of the search modal and search results pane.

- When the search modal is opened, uses `backdrop-filter` to blur the background instead of graying it out
- Removes loading spinner from search box and results pane (an overlay shown on loading was causing an unsightly flash
- Lowers debounce for the API call to Swifttype from `1500` to `200`, eliminating a lengthy delay for the user
- Removes a default `box-shadow` on the text input which was causing the input to appear as a few pixels less wide than the results pane... now both have the same box-shadow
- Styling updates: text in the input and the results are smaller, highlighted words appear in bold, underlined, and a different color

Test_cases

## How to test

Run `dr-ui` locally, use the search component to try out some search strings.  Expect highlighted words in the results to be blue, expect results to update without seeing a loading spinner.  

## Before merge

### DIFF
--- a/src/components/search/search-box.js
+++ b/src/components/search/search-box.js
@@ -150,7 +150,7 @@ export class SearchBox extends React.PureComponent {
     }
     function handleInputChange(newValue) {
       if (searchTerm === newValue) return;
-      setSearchTerm(newValue, { debounce: 1500 });
+      setSearchTerm(newValue, { debounce: 200 });
       // track query
       if (window && window.analytics) {
         analytics.track(segmentTrackEvent, {
@@ -192,11 +192,6 @@ export class SearchBox extends React.PureComponent {
               )}
               {(isOpen || resultsOnly) && searchTerm && (
                 <React.Fragment>
-                  {isLoading && (
-                    <div className="w-full h360 bg-lighten75 absolute z5">
-                      <div className="loading mx-auto mt60" />
-                    </div>
-                  )}
                   <div
                     className={classnames(
                       'color-text round mt3 bg-white w-full align-l',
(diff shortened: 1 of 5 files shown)


src/components/search/search-result.js
```diff
@@ -58,7 +58,26 @@ class SearchResult extends React.PureComponent {
...
+                      <em
+                        key={i}
+                        className="txt-underline txt-bold"
+                        style={{
+                          color: '#4264fb'
+                        }}
+                      >
+                        {' '}
+                        {node.children[0].data}
+                      </em>
```
VLuisa, line 75: @chriswhong I love the idea of highlighting the relevant terms in the search results! I feel like the current styling makes them look more like links rather than matches - what do you think about using the pattern of highlighting the matching search term instead?

Something like this:
image


```suggestion
                      <mark
                        key={i}
                        className="txt-bold bg-blue-light color-blue-dark"
                      >
                        {' '}
                        {node.children[0].data}
                      </mark>
```

src/components/search/search-box.js
```diff
@@ -150,7 +150,7 @@ export class SearchBox extends React.PureComponent {
     }
     function handleInputChange(newValue) {
       if (searchTerm === newValue) return;
-      setSearchTerm(newValue, { debounce: 1500 });
+      setSearchTerm(newValue, { debounce: 200 });
```
VLuisa, line 153: This does feel a lot nicer!

VLuisa (commented):
Just took a look! Overall I think these are great UX improvements, the smaller text size shows more results at once which is very helpful and the search matching is fantastic. I'm not convinced that the blurred out background is necessarily better than the grayed out version, but I may be biased because we don't use the blurred out pattern in Studio - but it's also not harmful, so up to you!

Left one suggestion for the styling of the search matches - feel free to take it or leave it! If you're not a fan of the highlight state, then I'd say to at least get rid of the underline styling because that makes the search term look too much like a link.

src/components/search/search-box.js
chriswhong: 😎

PR #553, opened 2023-02-28
fix search modal appearing on page navigation

chriswhong:
Fixes a bug where the search modal would open on page navigation after it had already been loaded and dismissed once.

https://mapbox.atlassian.net/jira/software/c/projects/MARKETING/boards/254?modal=detail&selectedIssue=MARKETING-1627


Countdown369 (approved):
LGTM

PR #555, opened 2023-03-15
bump mr-ui to 2.4.0, fix ariaLable console warning

chriswhong:
- Bumps `mr-ui` to the latest version `2.4.0` (no changes are introduced, this is just to stay current)

- Fixes the `ariaLabel` property warning which has been showing up on docs sites via `PageLayout` (the offending component was `BackToTopButton, which included an `ariaLabel` prop that should have been `aria-label`.  This is apparently an exception to the rule of using camelcase for html attributes in react.


Countdown369 (approved):
No more really, really long warnings sounds good to me! 🚀

PR #556, opened 2023-03-15
Fix cta layout

chriswhong:
Merge after #555 

- Adds some additional markup to create a better layout for the two CTA buttons (Sign up and Discord) when appearing at the bottom of a wide layout.
- Instead of showing up stacked, with lots of empty space, they now appear as 50% width tiles with gutter.

image


Countdown369 (commented):
Should the commit https://github.com/mapbox/dr-ui/pull/556/commits/68bb1ff1d07096f1cc983218b48de9cce90a20be be in this PR? I see that these are the changes from https://github.com/mapbox/dr-ui/pull/555. If this is fine, I'll approve (the `components` updates seem alright)

PR #557, opened 2023-03-22
update engines to node 18 and npm 9

chriswhong:
Updates dr-ui to use node 18/npm 9


Countdown369 (approved):
Looks like a standard node 18 update to me, 👍

PR #558, opened 2023-04-19
css and layout tweaks for the CTA buttons

chriswhong:
CSS and Layout Tweaks to fix the CTA buttons.

Mostly fixes the layout where they are showing as 1/2 width on small screens, but also fixes some bad spacing going on in other areas.

Geocoding_v6_API_Playground___Developer_Playgrounds___Mapbox


Countdown369 (approved):
LGTM 🙏

PR #570, opened 2023-05-10
restore babel present-env options

chriswhong:
The updates were causing an esm-style import statement to be added to the top of a cjs file


Countdown369 (approved):
👍

PR #574, opened 2023-05-15
Update 5.2.3

Countdown369:
Takes care of https://mapbox.atlassian.net/jira/software/c/projects/MARKETING/boards/254?modal=detail&selectedIssue=MARKETING-2179.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.2.3
+
+- add min-width:0 to Note component
+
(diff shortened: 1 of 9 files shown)


chriswhong (approved):
Approved, however in the changelog we should explain what the fix does.  e.g "styling fix to `Note` component to prevent text overflow"

chriswhong (changes requested):
This PR should only contain the code changes.  Once it is merged, then you do the changelog commit and bump the version number on the `main` branch

PR #595, opened 2023-08-08
Create Plugins component

Countdown369:
Create a Plugins component to be consumed by `mapbox-gl-js-docs`, `ios-sdk`, `android-docs`, and other unique "SDK"s that have plugins to be listed on the docs.


chriswhong (approved):
🚀

PR #600, opened 2023-08-09
improve the bulk pr script

chriswhong:
Improves `update-dr-ui-all.sh` to iterate over a list of repositories located in `scripts/target-repositories.json` and call the pr script for each one (instead of repeating hard-coded copies of the pr script)


Countdown369 (approved):
_Efficient._ LGTM

PR #605, opened 2023-09-13
update video modal

chriswhong:
- Reduces the modal width to 65vw instead of 80, which is more reasonable to keep the video in view when viewing the docs on a laptop.
- Refactors the styling to use the previous approach to maintaining the aspect ratio (uses % based padding so that the height does not need to be defined anywhere)


AlexanderBelokon (approved):
👍

PR #607, opened 2023-09-18
remove calls to mr-ui getWindow util

chriswhong:
Removes calls to `mr-ui` `getWindow()` util which causes issues when rendering SSR pages and is being removed from `mr-ui` in https://github.com/mapbox/mr-ui/pull/216


AlexanderBelokon (approved):
LGTM! The [original util](https://github.com/mapbox/mr-ui/blob/6950b90f84f852740b6d16c00c8af793e2f3a8f4/src/components/utils/get-window.ts#L4) doesn't do that much anyway...

PR #608, opened 2023-09-22
add IndexCard, helpCardData, build a card mode for OverviewHeader

chriswhong:
1. Adds `IndexCard`, formerly known as `DocsCard` and used on the docs landing page for each product.  Because we are now using it in two places (it will be used in the `help` docs) this brings it into dr-ui for re-use.
2. Adds an export called `helpCardData` under `/helpers`, which consolidates the colors and icons used in the various help sections.  These need to be in dr-ui because they are used on the docs landing page and in various places in the help docs.
3. Updates OverviewHeader to include `card` and `highlightColor` properties, this new Card-style design will be used in the help docs and replaces the older Overviewheader (solid color background + vector art):
image


AlexanderBelokon (changes requested):
The card looks great! 😍

One thing I would like to improve in the new code though, would be to rely on utils provided by Assembly a bit more and avoid setting literal `style` as much as possible.

Sizing utils have a (6px-based) scale for a reason, and slightly off sizes (like 28px instead of 30px or 4px instead of 3px/6px) make it harder to keep various elements neatly aligned, and generally makes things look less tidy and harder to support once these custom sizes accumulate in the codebase

src/components/overview-header/overview-header.js
```diff
@@ -117,24 +129,44 @@ class OverviewHeader extends React.PureComponent {
...
+                {title}
+                {tag && this.buildTag(this.props)}
+              </h1>
+              {titleIcon && (
+                <div
+                  className="mr18 relative"
+                  style={{
+                    height: 36,
+                    width: 36
+                  }}
```
AlexanderBelokon, line 159:
```suggestion
                  className="mr18 w36 h36 relative"
```

src/components/index-card/index-card.js
```diff
@@ -0,0 +1,187 @@
...
+        'px18 py18 flex flex--column': !isLinkList
+      })}
+    >
+      <div
+        className={classnames('flex flex--center-cross', {
+          mb12: icon
+        })}
+      >
+        {icon && (
+          <div className="mr12" style={{ height: 28, width: 28 }}>
```
AlexanderBelokon, line 70:
```suggestion
          <div className="mr12 w30 h30">
```

src/components/overview-header/overview-header.js
```diff
@@ -148,6 +180,25 @@ class OverviewHeader extends React.PureComponent {
...
+    if (card) {
+      return (
+        <div
+          className="flex overflow-hidden round-bold mt18 mb30"
+          style={{ boxShadow: cardBoxShadow }}
+        >
+          {card && (
+            <div
+              className={`flex-child-no-shrink bg-${highlightColor}`}
+              style={{ width: 7 }}
```
AlexanderBelokon, line 193:
```suggestion
              className={`w6 flex-child-no-shrink bg-${highlightColor}`}
```

AlexanderBelokon (approved):

src/components/overview-header/overview-header.js
chriswhong: Same, even one pixel changes the presentation of this highlighting bar and I think it's acceptable to leave a style prop here.

PR #611, opened 2023-09-27
style index card title/tag for proper vertical spacing when wrapped

chriswhong:
Updates the assembly classes used on the title in `IndexCard`, adding 6 
pixels of spacing when the content wraps to mutliple lines.
Before

image

After
image


mbxsocialimpact (approved):
lgtm - can't really tell the before/after difference, but I believe you ;)

PR #612, opened 2023-09-27
update search modal background to match mr-ui/modal

chriswhong:
Updates the search modal to use the same styling as `mr-ui/modal` for 
consist UX across the site. (e.g. `VideoModal` uses `mr-ui/modal` and we 
don't want two different modal experiences)


mbxsocialimpact (approved):
lgtm!

PR #613, opened 2023-10-06
export cardboxshadow from index-card

chriswhong:
Exports `cardBoxShadow` for use in other components and consuming docs sites.


chriswhong:
This gets removed during build, not relevant.

PR #620, opened 2023-10-25
Andrew/add playground type To Feedback Component

AndrewSepic:
- This change adds some intelligence to the `<ContentWrapper>` component to check props for the `layoutConfig.contentType` to determine if the location in which the component is being rendered is an example page or a playground page.
- If it is either of these types, the value is passed into the `type` prop of the `<Feedback>` component.  If it's not either of these values, an undefined value is passed into the type and `<Feedback>`'s default prop type of `page` is used.  

## How to test

- Ensure that your local Dr-UI repo is connected (via `npx link`) to the playgrounds repo, and the GL-JS-docs repo.  
- Pull down this branch, run `npm run build` in dr-ui
- Start the connected playground repo and check the output of the feedback module on a playground page to ensure it reads 'Was this playground helpful?`.  
- Start the connected  gl-js-docs repo and load an example page to ensure the output on the feedback module is ' Was this example helpful?'
- Check other locations in gl-js-docs live preview to ensure 'Was this page helpful?' and 'Was this section on ______ helpful?' are also returned in the appropriate places.

### DIFF
--- a/src/components/feedback/feedback.js
+++ b/src/components/feedback/feedback.js
@@ -27,6 +27,7 @@ FeedbackWrapper.propTypes = {
 };
 
 const Feedback = (props) => {
+  console.log('props are', props);
   const { type } = props;
 
   const [user, setUser] = useState(undefined);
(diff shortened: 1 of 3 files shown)


chriswhong (approved):
LGTM!

PR #624, opened 2023-11-01
update dependencies

chriswhong:
- Updates a few dependencies
- Refactors `build-docs-data` to use new async methods of `prettier` which were introduced in a downstream dependency
- Uses a beta release of `copyeditor` which uses `.copyeditorrc` instead of a local `.dict` file

- Remaining vulnerabilities are related to batfish


AndrewSepic (approved):
Looks good with the webpack loader fix!

PR #634, opened 2023-11-29
Andrew/add faq type

AndrewSepic:
This PR introduces a new content type for Mapbox docs. The 'FAQ' content type.  

- It adds a new Icon SVG and content Type to `help-card-data.js` 
- Changes the order of the content types to put troubleshooting last.
- It also updates `mbx-assembly-docs` to unlock a new color for this new content type.

**EDIT** 
- This PR also instroduces hover state and styling for the `card.js` component.


package.json
```diff
@@ -84,6 +84,7 @@
     "@elastic/search-ui-site-search-connector": "1.7.0",
     "@mapbox/mapbox-gl-supported": "^2.0.1",
     "@mapbox/mr-ui": "^2.7.0",
+    "@mapbox/remark-lint-mapbox": "^4.1.0",
```
chriswhong, line 87: Does this need to be explicitly added here?

chriswhong (approved):
Just one question about `remark-lint-mapbox`.  Looks good!

package.json
AndrewSepic: No, I believe this was left over from my forced update to `remark-lint-mapbox` in order to have access to the new content type.  This packages is already a dependency of `remark-config-docs` but rather than wait for dependabot to push an updated release (24 hrs?) I ran `npm i @mapbox/remark-lint-mapbox@latest --force` and I believe that's what added this. I was following the [release instructions here](https://github.com/mapbox/remark-lint-mapbox#how-to-release). 

I've removed this in the my most recent commit and `remark-config-docs` should have the most recent version (if the former release process is still working)

src/components/card/card.js
```diff
@@ -31,28 +45,52 @@ class Card extends React.PureComponent {
...
+                  marginLeft: isHovered ? '15px' : '10px',
+                  opacity: isHovered ? '1' : '0'
+                }}
+                width="11"
+                height="11"
+                viewBox="0 0 11 11"
+                fill="none"
+                xmlns="http://www.w3.org/2000/svg"
+              >
+                <path
```
chriswhong, line 77: We should move this arrow icon out into a component so it is re-usable, and so it's clearer what it is in the code.

chriswhong (changes requested):

AndrewSepic:
@chriswhong   Updates on this last PR.  The changes needed to make the `RightArrow` icon available for use within the `pkg/components` folder ofg `dr-ui` may not be the best long term solution.  Now an icon has to be considered a fundamental part of `dr-ui` and have an example and be included in `docs/src/categories.json`.  I think the best approach here may be to keep this SVG inline within the `card.js` or put in place a a new folder strcuture & build process for including icon components in `dr-ui` without all the overhead that it has now.  I want to leave this as is for now, but have created a [Jira ticket](https://mapbox.atlassian.net/browse/MARKETING-2974) to keep track of this.  

Lastly,  I had issues trying to work around the `react/jsx-no-bind` eslint rule that does not want arrow functions used as component props.  I am not as competent with class based components and struggled to write a non arrow function and have it bind to the component instance and work.  

I'm happy to pair on that with you to help me understand, or keep the `src/components/card/card.js` as is, using `//eslint-ignore`

chriswhong (approved):
LGTM, added a commit to avoid the arrow function/eslint ignore issue in card.js

